PR #4, opened 2023-12-11
Expose OIDC config parameters

arvchristos:
This MR exposes OIDC configuration parameters in the default config entrypoint script and the env variables. Essentially, `misp-docker` can now easily work with external SSO providers, a requirement for our use-case.

Settings are configured according to https://github.com/MISP/MISP/blob/2.4/app/Plugin/OidcAuth/README.md

By default, OIDC is disabled but can be enabled with the env variable `OIDC_ENABLE`.


ostefano:
Thank you @arvchristos , I like this.

We are using OIDC as well, so might take this as an opportunity to merge some additonal logic.
I will iterate later today.

core/files/configure_misp.sh
```diff
@@ -93,6 +103,33 @@ GPGEOF
     sudo -u www-data /var/www/MISP/app/Console/cake Admin setSetting -q "GnuPG.binary" "$(which gpg)"
 }
 
+set_up_oidc() {
+    if [[ "$OIDC_ENABLE" != "true" ]]; then
+        echo "... OIDC authentication disabled"
+        return
+    fi
+
+    sudo -u www-data php /var/www/MISP/tests/modify_config.php modify "{
```
ostefano, line 107: Note that this is not idempotent (can't recall the exact behavior :S).

Best fix, is to reset the value before setting it.

```
sudo -u www-data php /var/www/MISP/tests/modify_config.php modify "{
            \"Security\": {
                \"auth\": \"\"
            }
}" > /dev/null 
```

template.env
```diff
@@ -87,3 +87,12 @@ SYNCSERVERS_1_KEY=
 
 # Disable IPv6 completely (this setting will persist until the container is removed)
 # DISABLE_IPV6=true
+
+# Enable OIDC authentication, according to https://github.com/MISP/MISP/blob/2.4/app/Plugin/OidcAuth/README.md
+# OIDC_ENABLE=true
+# OIDC_PROVIDER_URL=
+# OIDC_CLIENT_ID=
+# OIDC_CLIENT_SECRET=
+# OIDC_ROLES_PROPERTY=
```
ostefano: I would add some example re the data type

template.env
```diff
@@ -87,3 +87,12 @@ SYNCSERVERS_1_KEY=
...
+# OIDC_ROLES_MAPPING=
```
ostefano: Same thing here.

core/files/configure_misp.sh
```diff
@@ -6,10 +6,20 @@ source /rest_client.sh
 [ -z "$GPG_PASSPHRASE" ] && GPG_PASSPHRASE="passphrase"
 [ -z "$REDIS_FQDN" ] && REDIS_FQDN="redis"
 [ -z "$MISP_MODULES_FQDN" ] && MISP_MODULES_FQDN="http://misp-modules"
+[ -z "$OIDC_PROVIDER_URL" ] && OIDC_PROVIDER_URL="test_provider"
```
ostefano: Not sure whether we want default values if not working

core/files/configure_misp.sh
arvchristos: I thought of this but this step is already happening at https://github.com/MISP/misp-docker/blob/526c47a6e8b92e1c761030b7a93a95c7b6c85bf3/core/files/configure_misp.sh#L192  which is by default executed and cannot be disabled with a flag.

core/files/configure_misp.sh
ostefano: Let me rephrase it. I would rather remove default values, and add some checks when setting up oidc (e.g., if not all values are set, error out).

arvchristos: Done and added a generic  utility which can be used for other configuration functions in the future

ostefano: Nice. I see the output is redirected to stderr. Should we leave it like that?

arvchristos: Actually no, you are totally right, fixed

template.env
arvchristos: Fixed

template.env
arvchristos: Fixed

ostefano:
LGTM

PR #36, opened 2024-03-29
Add healthchecks to docker-compose.yml

ftoppi:
Hello,

here's a small PR to replace redis with keydb due to redis license changes and to add healthchecks to services.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,6 +13,12 @@ services:
 
   redis:
     image: eqalpha/keydb:x86_64_v6.3.4
+    healthcheck:
+      test: keydb-cli ping || exit 1
+      interval: 2s
+      timeout: 1s
+      retries: 3
+      start_period: 30s
 
   db:
     # We use MariaDB because it supports ARM and has the expected collations
@@ -27,6 +33,12 @@ services:
       - mysql_data:/var/lib/mysql
     cap_add:
       - SYS_NICE  # CAP_SYS_NICE Prevent runaway mysql log
+    healthcheck:
+      test: mysqladmin --user=$$MYSQL_USER --password=$$MYSQL_PASSWORD status
+      interval: 2s
+      timeout: 1s
+      retries: 3
+      start_period: 30s
 
   misp-core:
     image: ghcr.io/misp/misp-docker/misp-core:latest
@@ -46,8 +58,16 @@ services:
           - PYPI_CYBOX_VERSION=${PYPI_CYBOX_VERSION}
           - PYPI_PYMISP_VERSION=${PYPI_PYMISP_VERSION}
     depends_on:
-      - redis
-      - db
+      redis:
+        condition: service_healthy
+      db:
+        condition: service_healthy
+    healthcheck:
+      test: curl -ks https://localhost/users/login > /dev/null || exit 1
+      interval: 2s
+      timeout: 1s
+      retries: 3
+      start_period: 30s
     ports:
       - "80:80"
       - "443:443"
@@ -126,7 +146,8 @@ services:
     environment:
       - "REDIS_BACKEND=redis"
     depends_on:
-      - redis
+      redis:
+        condition: service_healthy
 
 volumes:
     mysql_data:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -12,7 +12,7 @@ services:
       - "SMARTHOST_ALIASES=${SMARTHOST_ALIASES}"
 
   redis:
-    image: redis:7.2
+    image: eqalpha/keydb:x86_64_v6.3.4
```
ostefano: This image is x64 only while the Redis image is ARM as well. Would it be possible to remove the arch modifier here?

docker-compose.yml
```diff
@@ -46,8 +58,16 @@ services:
...
           - PYPI_PYMISP_VERSION=${PYPI_PYMISP_VERSION}
     depends_on:
-      - redis
-      - db
+      redis:
+        condition: service_healthy
+      db:
+        condition: service_healthy
+    healthcheck:
+      test: curl -ks https://localhost/users/login > /dev/null || exit 1
```
ostefano, line 66: Is this check run inside the container? In other words, Is `localhost` the correct setting regardless of our `BASE_URL`?

docker-compose.yml
ostefano: Note that unless MISP changes the default key-value storage, we should only document this alternative in the README.md rather than modify the `docker-compose.yml` file

ftoppi: Fine by me. All their images are arch-specific except `eqalpha/keydb:latest` which is both for arm64 and x86_64.

docker-compose.yml
ftoppi: Yes, it's run inside the container.
You can see below I don't expose ports in my stack and it's still healthy:

```
$ docker compose ps 2> /dev/null
NAME                  IMAGE                                          COMMAND                  SERVICE        CREATED        STATUS                  PORTS
misp-db-1             mariadb:10.11                                  "docker-entrypoint.s…"   db             43 hours ago   Up 43 hours (healthy)   3306/tcp
misp-mail-1           ixdotai/smtp                                   "/bin/entrypoint.sh …"   mail           43 hours ago   Up 43 hours             25/tcp
misp-misp-core-1      ghcr.io/misp/misp-docker/misp-core:latest      "/entrypoint.sh"         misp-core      42 hours ago   Up 42 hours (healthy)
misp-misp-modules-1   ghcr.io/misp/misp-docker/misp-modules:latest   "/usr/local/bin/misp…"   misp-modules   43 hours ago   Up 43 hours
misp-redis-1          eqalpha/keydb:x86_64_v6.3.4                    "docker-entrypoint.s…"   redis          43 hours ago   Up 43 hours (healthy)   6379/tcp
```

docker-compose.yml
ostefano: We should also reference https://github.com/valkey-io/valkey if a docker image is available.

docker-compose.yml
```diff
@@ -12,7 +12,13 @@ services:
       - "SMARTHOST_ALIASES=${SMARTHOST_ALIASES}"
 
   redis:
-    image: redis:7.2
+    image: eqalpha/keydb:x86_64_v6.3.4
+    healthcheck:
+      test: keydb-cli ping || exit 1
```
ostefano: What is the alternative command if were to use Redis here?

ostefano (changes requested):
This is good stuff, but I think it's too early to replace redis.

Can we add the healthchecks first, and only when better docker images are published, move to a new key value store?

Alternatively, it should be easy to just document the alternatives available in the README.

docker-compose.yml
ftoppi: redis-cli ping

ostefano (changes requested):
Will merge if changes to the underlying image are removed.

Also, note that we decided to migrate to valkey, so healthcheck commands might need updating.

ftoppi:
> Will merge if changes to the underlying image are removed.
> 
> Also, note that we decided to migrate to valkey, so healthcheck commands might need updating.

I figured everyone would move to valkey but it wasn't available at the time.

This is better?

ostefano (approved):

ostefano:
Looks good.

Can you rebase and squash? Otherwise I will do it when merging (will wait for next MISP release).

PR #39, opened 2024-04-05
Add AadAuth support in configure_misp.sh

### LINKED ISSUE #34: AAD Auth not working
Hello, 

I'm using CoolAcid container version on my production environment, and when I start the migration to this container, I've noticed the AAD authentication is not working or not supported anymore. 

I have the put all settings on config.php file and add cake module to bootstrap.php file also.

shieldsurge:
Add support for enabling Azure AD (aka Entra) authentication via the configure_misp.sh script. The function to enable Azure AD auth mirrors the code in the existing functions to enable OIDC and LDAP auth.

Environment variables are provided to configure available options in the AadAuth plugin: https://github.com/MISP/MISP/tree/2.4/app/Plugin/AadAuth

This commit addresses issue "AAD Auth not working" #34. https://github.com/MISP/misp-docker/issues/34

### DIFF
--- a/core/files/configure_misp.sh
+++ b/core/files/configure_misp.sh
@@ -157,6 +157,61 @@ set_up_ldap() {
     }" > /dev/null
 }
 
+set_up_aad() {
+    if [[ "$AAD_ENABLE" != "true" ]]; then
+        echo "... Entra (AzureAD) authentication disabled"
+        return
+    fi
+
+    # Check required variables
+    check_env_vars AAD_CLIENT_ID AAD_TENANT_ID AAD_CLIENT_SECRET BASE_URL
+
+    # Configure unset optional AAD environment variables to default values
+    [ -z "$AAD_REDIRECT_URI" ] && AAD_REDIRECT_URI="${BASE_URL}/users/login"
+    [ -z "$AAD_PROVIDER" ] && AAD_PROVIDER=https://login.microsoftonline.com/
+    [ -z "$AAD_PROVIDER_USER" ] && AAD_PROVIDER_USER=https://graph.microsoft.com/
+    [ -z "$AAD_MISP_USER" ] && AAD_MISP_USER="Misp Users"
+    [ -z "$AAD_MISP_ORGADMIN" ] && AAD_MISP_ORGADMIN="Misp Org Admins"
+    [ -z "$AAD_MISP_SITEADMIN" ] && AAD_MISP_SITEADMIN="Misp Site Admins"
+    [ -z "$AAD_CHECK_GROUPS" ] && AAD_CHECK_GROUPS=false
+
+    # Note: Not necessary to edit bootstrap.php to load AadAuth Cake plugin because 
+    # existing loadAll() call in bootstrap.php already loads all available Cake plugins
+
+    # Set auth mechanism to AAD in config.php file
+    sudo -u www-data php /var/www/MISP/tests/modify_config.php modify "{
+        \"Security\": {
+            \"auth\": [\"AadAuth.AadAuthenticate\"]
+        }
+    }" > /dev/null
+
+    echo "hello"
+    # Configure AAD auth settings from environment variables in config.php file
+    sudo -u www-data php /var/www/MISP/tests/modify_config.php modify "{
+        \"AadAuth\": {
+            \"client_id\": \"${AAD_CLIENT_ID}\",
+            \"ad_tenant\": \"${AAD_TENANT_ID}\",
+            \"client_secret\": \"${AAD_CLIENT_SECRET}\",
+            \"redirect_uri\": \"${AAD_REDIRECT_URI}\",
+            \"auth_provider\": \"${AAD_PROVIDER}\",
+            \"auth_provider_user\": \"${AAD_PROVIDER_USER}\",
+            \"misp_user\": \"${AAD_MISP_USER}\",
+            \"misp_orgadmin\": \"${AAD_MISP_ORGADMIN}\",
+            \"misp_siteadmin\": \"${AAD_MISP_SITEADMIN}\",
+            \"check_ad_groups\": ${AAD_CHECK_GROUPS}
+        }
+    }" > /dev/null
+
+    # Disable self-management, username change, and password change to prevent users from circumventing AAD login flow
+    # Recommended per https://github.com/MISP/MISP/blob/2.4/app/Plugin/AadAuth/README.md
+    sudo -u www-data /var/www/MISP/app/Console/cake Admin setSetting -q "MISP.disableUserSelfManagement" true
+    sudo -u www-data /var/www/MISP/app/Console/cake Admin setSetting -q "MISP.disable_user_login_change" true
+    sudo -u www-data /var/www/MISP/app/Console/cake Admin setSetting -q "MISP.disable_user_password_change" true
+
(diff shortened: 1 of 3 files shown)


core/files/configure_misp.sh
```diff
@@ -157,6 +157,60 @@ set_up_ldap() {
...
+    if [[ "$AAD_ENABLE" != "true" ]]; then
+        echo "... Entra (AzureAD) authentication disabled"
+        return
+    fi
+
+    # Check required variables
+    check_env_vars AAD_CLIENT_ID AAD_TENANT_ID AAD_CLIENT_SECRET BASE_URL
+
+    # Configure unset optional AAD environment variables to default values
+    [ -z "$AAD_REDIRECT_URI" ] && AAD_REDIRECT_URI="${BASE_URL}/users/login"
```
ostefano: Other auth mechanisms do not set default but leave suggested values in the `template.env` file. AAD should do the same since MS is not the only implementation (let me know if that is instead the case).

core/files/configure_misp.sh
```diff
@@ -157,6 +157,60 @@ set_up_ldap() {
...
 }
 
+set_up_aad() {
+    if [[ "$AAD_ENABLE" != "true" ]]; then
+        echo "... Entra (AzureAD) authentication disabled"
+        return
+    fi
+
+    # Check required variables
+    check_env_vars AAD_CLIENT_ID AAD_TENANT_ID AAD_CLIENT_SECRET BASE_URL
```
ostefano: No need to check BASE_URL, I would rather check all other vars

ostefano (changes requested):
Thank you for the PR.

We should also update `docker-compose.yml` and `template.env`

core/files/configure_misp.sh
shieldsurge: The goal here was to provide a few default variable values to allow the administrator to specify a minimal set of environment variables for AAD auth while still preserving their ability to override any of the default variables by setting the environment variables explicitly.

For example, the [AadAuth plugin readme](https://github.com/MISP/MISP/blob/2.4/app/Plugin/AadAuth/README.md) provides example values for the `auth_provider` URI and the `auth_provider_user` URI, and comments "No change required" for both. I believe these values are the same for everybody, so the idea was to set them to the default values unless explicitly overridden in environment variables.

There's also optional functionality in the AadAuth plugin to have MISP require the authenticating end user to be a member of one of three administrator-defined Azure AD groups to allow login. This functionality isn't very useful at the moment because there's no mechanism to map roles in MISP to Azure AD groups as far as I can tell. Therefore, I made the default value for this functionality off (`AAD_CHECK_GROUPS=false`), and the three Azure AD group names are just set to the example values specified in the AadAuth plugin readme page.

`AAD_REDIRECT_URI` is going to be `"${BASE_URL}/users/login"` in 99% of implementations, so I specified it as the default which can be overridden by setting the env variable.

All that being said, I don't feel strongly about this issue. If the desired practice is to require the administrator to explicitly specify all possible AAD environment variables, just let me know and I'll adjust the code accordingly.

@ostefano for awareness.

core/files/configure_misp.sh
shieldsurge: I omitted some of the environment variables from the `check_env_vars` call because I set them to default values if not specified, per the discussion below. If that's not the direction we want to go in, then I will add those variables to the `check_env_vars` call to require them to be set explicitly.

shieldsurge:
> Thank you for the PR.
> 
> We should also update `docker-compose.yml` and `template.env`

`docker-compose.yml` and `template.env` have been updated.

ostefano:
> > Thank you for the PR.
> > We should also update `docker-compose.yml` and `template.env`
> 
> `docker-compose.yml` and `template.env` have been updated.

Thank you,  but defaults are still set in the configure_misp file. Can we remove it from there so all auth mechanisms are consistent with each other?

shieldsurge:
> > > Thank you for the PR.
> > > We should also update `docker-compose.yml` and `template.env`
> > 
> > 
> > `docker-compose.yml` and `template.env` have been updated.
> 
> Thank you, but defaults are still set in the configure_misp file. Can we remove it from there so all auth mechanisms are consistent with each other?

Understood, assuming you saw my rationale for default values in the review section. I went ahead and removed the default values from `core/files/configure_misp.sh`. It now requires all AAD env vars to be explicitly set when using AAD to be consistent with the approach used by the other auth mechanisms.

ostefano:
> > > > Thank you for the PR.
> > > > We should also update `docker-compose.yml` and `template.env`
> > > 
> > > 
> > > `docker-compose.yml` and `template.env` have been updated.
> > 
> > 
> > Thank you, but defaults are still set in the configure_misp file. Can we remove it from there so all auth mechanisms are consistent with each other?
> 
> Understood, assuming you saw my rationale for default values in the review section. I went ahead and removed the default values from `core/files/configure_misp.sh`. It now requires all AAD env vars to be explicitly set when using AAD to be consistent with the approach used by the other auth mechanisms.

Mmm, I don't see any reply in the review section. Did I miss your comments?

shieldsurge:
> > > > > Thank you for the PR.
> > > > > We should also update `docker-compose.yml` and `template.env`
> > > > 
> > > > 
> > > > `docker-compose.yml` and `template.env` have been updated.
> > > 
> > > 
> > > Thank you, but defaults are still set in the configure_misp file. Can we remove it from there so all auth mechanisms are consistent with each other?
> > 
> > 
> > Understood, assuming you saw my rationale for default values in the review section. I went ahead and removed the default values from `core/files/configure_misp.sh`. It now requires all AAD env vars to be explicitly set when using AAD to be consistent with the approach used by the other auth mechanisms.
> 
> Mmm, I don't see any reply in the review section. Did I miss your comments?

I can't seem to figure out how to link to the comment, but here's what I wrote:

The goal here was to provide a few default variable values to allow the administrator to specify a minimal set of environment variables for AAD auth while still preserving their ability to override any of the default variables by setting the environment variables explicitly.

For example, the [AadAuth plugin readme](https://github.com/MISP/MISP/blob/2.4/app/Plugin/AadAuth/README.md) provides example values for the `auth_provider` URI and the `auth_provider_user` URI, and comments "No change required" for both. I believe these values are the same for everybody, so the idea was to set them to the default values unless explicitly overridden in environment variables.

There's also optional functionality in the AadAuth plugin to have MISP require the authenticating end user to be a member of one of three administrator-defined Azure AD groups to allow login. This functionality isn't very useful at the moment because there's no mechanism to map roles in MISP to Azure AD groups as far as I can tell. Therefore, I made the default value for this functionality off (`AAD_CHECK_GROUPS=false`), and the three Azure AD group names are just set to the example values specified in the AadAuth plugin readme page.

`AAD_REDIRECT_URI` is going to be `"${BASE_URL}/users/login"` in 99% of implementations, so I specified it as the default which can be overridden by setting the env variable.

All that being said, I don't feel strongly about this issue. If the desired practice is to require the administrator to explicitly specify all possible AAD environment variables, just let me know and I'll adjust the code accordingly.

shieldsurge:
> Mmm, I don't see any reply in the review section. Did I miss your comments?

It was because I never submitted the review 🤦
You should be able to see the comments now.

ostefano (approved):
I will merge this, but add some changes afterwards

PR #41, opened 2024-04-09
Fix #40 Add exemption to entrypoint_nginx for taxonomies

### LINKED ISSUE #40: Custom taxonomies get overwritten on startup
Core Build: v2.4.188

**Description**
When starting the core container the `files/taxonomies` directory is overwritten by `entrypoint_nginx.sh`. This seems to prevent the user from provisioning custom taxonomies to MISP. 

**Related issues**
#31
#16 
#17 

Can this be fixed by applying a similar exception as used in the related issues, or am I misunderstanding how taxonomies should be provided?

DaanWillems:
### What does it do?
The change fixes https://github.com/MISP/misp-docker/issues/40

### Expected behavior
The taxonomies should be persistent after restarting or updating the core container

### DIFF
--- a/core/files/entrypoint_nginx.sh
+++ b/core/files/entrypoint_nginx.sh
@@ -148,7 +148,7 @@ EOT
 
 update_misp_data_files(){
     for DIR in $(ls /var/www/MISP/app/files.dist); do
-        if [ "$DIR" = "certs" ] || [ "$DIR" = "img" ] ; then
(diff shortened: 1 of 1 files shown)


ostefano:
Thank you @DaanWillems . Since this is the third time it happens, can you investigate whether we should do the same for other directories?

ostefano (approved):

PR #49, opened 2024-04-19
Allow Redis over TLS

### LINKED ISSUE #47: Redis over TLS incompatible due to hardcoded tcp:// in entrypoint
### Context
Currently it is possible to use **Redis over TLS** by appending `tls://` to the `REDIS_FQDN` environment variable, which even if not present in the `template.env` file, is available in the code:
https://github.com/MISP/misp-docker/blob/67d4a0ea326155189c7549b451a1e55a213e2ab4/core/files/configure_misp.sh#L8
Therefore, by setting up this variable in the `.env` file, it is possible to communicate to an external Redis host, which would be via TLS if it's set up as `REDIS_FQDN: "tls://redis.example.com"`
### Error
Unfortunately, the `entrypoint_fpm.sh` file hardcodes the `tcp://` prefix in the FQDN:
https://github.com/MISP/misp-docker/blob/67d4a0ea326155189c7549b451a1e55a213e2ab4/core/files/entrypoint_fpm.sh#L22
Which makes PHP fail to connect to Redis, since the endpoint now looks like `tcp://tls://redis.example.com`.

christianmg99:
### What does it do?
It fixes #47 by prepending `tcp://` only when not present in `$REDIS_FQDN` already.

### DIFF
--- a/core/files/entrypoint_fpm.sh
+++ b/core/files/entrypoint_fpm.sh
@@ -19,7 +19,7 @@ change_php_vars() {
         sed -i "s/upload_max_filesize = .*/upload_max_filesize = 50M/" "$FILE"
         sed -i "s/post_max_size = .*/post_max_size = 50M/" "$FILE"
         sed -i "s/session.save_handler = .*/session.save_handler = redis/" "$FILE"
-        sed -i "s|.*session.save_path = .*|session.save_path = 'tcp://${REDIS_FQDN}:6379'|" "$FILE"
+        sed -i "s|.*session.save_path =.*|session.save_path = '$(echo $REDIS_FQDN | grep -E '^\w+://' || echo tcp://$REDIS_FQDN):6379'|" "$FILE"
     done
 }
 
--- a/core/files/entrypoint_fpm.sh
+++ b/core/files/entrypoint_fpm.sh
@@ -19,7 +19,7 @@ change_php_vars() {
         sed -i "s/upload_max_filesize = .*/upload_max_filesize = 50M/" "$FILE"
         sed -i "s/post_max_size = .*/post_max_size = 50M/" "$FILE"
         sed -i "s/session.save_handler = .*/session.save_handler = redis/" "$FILE"
-        sed -i "s|.*session.save_path =.*|session.save_path = '$(echo $REDIS_FQDN | grep -E '^\w+://' || echo tcp://$REDIS_FQDN):6379'|" "$FILE"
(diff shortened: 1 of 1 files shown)


ostefano:
@christianmg99  is tls://redis.example.com handled correctly by MISP itself? In other words, what about configure_misp.sh? Did you try building the image and looking at diagnostics?

ostefano (changes requested):
See comment. We need to make sure that `tls://` prefixed strings are correctly managed by MISP itself.

christianmg99:
By the time I found the replacement was in the entrypoint I had already read all Misp code regarding redis_host, and all I saw was mindful about the variable starting with tls:// such as in [zmq plugin](https://github.com/MISP/MISP/blob/471840ce33233a8d9db86766d007153f66727e8d/app/files/scripts/mispzmq/mispzmq.py#L91) and in [PHP-redis plugin](https://github.com/phpredis/phpredis/blob/62cf943fecc5182c6329b332df43cf28012cef55/README.md?plain=1#L241), I also looked into cakePHP and it has url prefix parsers as well. I currently use it like that in a custom deployment and it works fine.

christianmg99:
Anyways, as it is coded, you would need to prepend tls:// anyways for it to be used, so it's only for the ones who want to "force" tls use in the variable like me that will take place, everyone else will follow the standard tcp:// prepend.

core/files/entrypoint_fpm.sh
```diff
@@ -19,7 +19,7 @@ change_php_vars() {
         sed -i "s/upload_max_filesize = .*/upload_max_filesize = 50M/" "$FILE"
         sed -i "s/post_max_size = .*/post_max_size = 50M/" "$FILE"
         sed -i "s/session.save_handler = .*/session.save_handler = redis/" "$FILE"
-        sed -i "s|.*session.save_path = .*|session.save_path = 'tcp://${REDIS_FQDN}:6379'|" "$FILE"
```
ostefano, line 22: Why removing the space after `=`?

christianmg99: With the space there, if the space goes missing in the original file the replacement never takes place. It is more human readable with the space but facilitates some window for sed not catching on the string. Anyways if you want I can commit with the space back.

christianmg99: Or feel free to do so yourself, I believe I open the request allowing you to commit on top

ostefano: I am not the original author of these `sed` commands, so I would rather leave them as such or change them all.

christianmg99: Space is back :)

ostefano (approved):

PR #52, opened 2024-04-23
Support optional OIDC_ISSUER parameter

christianmg99:
Following https://github.com/MISP/MISP/pull/9694 merge, supporting new optional OIDC parameter.

From shell docs:
`${parameter:+word}`
`If parameter is null or unset, nothing is substituted, otherwise the expansion of word is substituted.`

I used this method since issuer should be null and not an empty string in case it's missing (Unsure about whether MISP understands an empty string as null value).

### DIFF
--- a/core/files/configure_misp.sh
+++ b/core/files/configure_misp.sh
@@ -103,6 +103,7 @@ set_up_oidc() {
     fi
 
     # Check required variables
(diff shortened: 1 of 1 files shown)


ostefano:
LGTM 👍

PR #54, opened 2024-04-29
Fix issue where searching for '.php' returns an error

UFOSmuggler:
If you search MISP for something ending in .php, the nginx location directive is too broad, and will match.  This results in nginx attempting to load the search path as a file.

Consider:
- Search for `something.php`
- The URL accessed becomes `/events/index/searcheventinfo:something.php`
- The MISP nginx location directive this passes through is predicated on the path`\.php$`

This will obviously match, and nginx attempt to load `searcheventinfo:something.php`, and this is what we see:

image

This commit causes the file path (`$fastcgi_script_name`) and subsequent path (`$fastcgi_path_info`) to be passed as separate parameters to fastcgi, so that any arbitrary path ending in .php is not interpreted as a script name.

I've been running this fix for months on my own misp-docker instances, and haven't had any issues.

### DIFF
--- a/core/files/etc/nginx/includes/misp
+++ b/core/files/etc/nginx/includes/misp
@@ -21,8 +21,11 @@ location / {
     try_files $uri $uri/ /index.php$is_args$query_string;
 }
 
-location ~ \.php$ {
+location ~ ^/[^/]+\.php(/|$) {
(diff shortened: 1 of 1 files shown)


ostefano:
LGTM 👍

PR #56, opened 2024-05-02
Allow REDIS Password and ZMQ Switch

christianmg99:
This code change allows for 2 new optional env variables REDIS_PASSWORD and ZMQ_ENABLE.
Current deployments without these variables aren't affected.


ostefano:
Not sure whether redis password is correctly loaded from docker-compose. <strike>Anyway ZMQ and redis password are two different concerns, so it would make sense to split the PR.</strike>

ostefano (changes requested):
LGTM, but we should have the env vars documented inside template.env, and picked up and forwarded to the container in the docker-compose.yml file as well.

ostefano:
> LGTM, but we should have the env vars documented inside template.env, and picked up and forwarded to the container in the docker-compose.yml file as well.

Also `misp-modules` needs to pick up the password as well.

ostefano:
This PR would solve https://github.com/MISP/misp-docker/issues/9

ostefano:
Broke as is, https://github.com/MISP/misp-docker/pull/107 supersedes it

PR #66, opened 2024-05-28
Alter apply_critical_fixes() to check that a setting has not already been set before forcing to safe default.

### LINKED ISSUE #63: apply_critical_fixes() in configure_misp.sh turns off actions plugin every time the misp-core container comes up
https://github.com/MISP/misp-docker/blob/master/core/files/configure_misp.sh#L264

I suspect this is forced off so that if the Action_services_url is incorrect, MISP doesn't drop its lunch.  However, it doesn't seem to cause an issue if set to something that doesn't exist.

Perhaps this isn't needed after all?  Otherwise, every time the container comes up we have to edit the config again to turn it back on.

Is there some other reason for this I am not aware of?

UFOSmuggler:
Fix for #63 

Uses cake console commands to get and set settings.
Works with both DB and config.php settings.

Unfortunately, MISP's serverSettings model does not seem to understand Security.auth, which is used for external auth purposes.  For this reason we have to do ugly kludge to check this array, then use the existing ugly kludge to set it as an empty array if it does not already exist.  For some reason, the previous script was setting it as "" instead of array(), but it is clearly expected to be an array.  Probably doesn't really matter though, and we leave it alone if already "" after this change.


ostefano:
What about `apply_optional_fixes`? Would it make sense to extend this mechanism to those settings?
Maybe by extracting the set logic, and making it a function accepting a JSON-like variable?

UFOSmuggler:
ok, generalising the method to do this but running into a few misp related issues.

also fixing up a few other minor things at the same time that have been annoying me.

give me a few days and i should have a sound method to deal with settings generally.

ostefano:
I like this approach. Do you think it would make sense to split config and data and store the JSON data outside the .sh file? That would improve readability/maintainability.

UFOSmuggler:
take a gander at that and let me know what you think.

ostefano:
Nice stuff! Would it make sense to merge `enforced` and non-enforced files, and add a key to each entry `enforced=False/True`?

Little nit: change of tag in docker-compose should be not pushed.

ostefano (changes requested):
See https://github.com/MISP/misp-docker/pull/66#issuecomment-2141853372

UFOSmuggler:
> Nice stuff! Would it make sense to merge `enforced` and non-enforced files, and add a key to each entry `enforced=False/True`?

I am of the opinion that the function of the two files is sufficiently different that they should remain separated for safety.  If you still believe that they should be merged, rather than an "enforced" boolean key, probably better to have "enforced" and "unenforced" object keys that contain their respective settings to apply.  still, i'd prefer them left as is to avoid unintentional foot shootings.

> 
> Little nit: change of tag in docker-compose should be not pushed.

rebased and pushed without this accident.

UFOSmuggler (commented):
See https://github.com/MISP/misp-docker/pull/66#issuecomment-2144033908

ostefano:
@UFOSmuggler followed up in gitter/matrix.

ostefano (changes requested):
As per discussion:

* add some doc in the readme (especially when, i.e., which scenarios, we should update the enforced / non_enforced JSON to add new settings)
* comment out code that is not used
* test the changeset against more complex configurations (OIDC/LDAP)
* test the changeset whether settings in DB / config are used
* change the file names so files containing the same settings are next to each other when sorted

UFOSmuggler:
* ~~add some doc in the readme (especially when, i.e., which scenarios, we should update the enforced / non_enforced JSON to add new settings)~~
* ~~comment out code that is not used~~
* ~~test the changeset against more complex configurations (OIDC/LDAP)~~
* ~~test the changeset whether settings in DB / config are used~~
* ~~change the file names so files containing the same settings are next to each other when sorted~~
* ~~Check custom certs still work~~

ostefano:
One thing I forgot to ask: what is the impact on users updating to a docker image featuring this new settings system? Is it something we can qualify/quantify?

UFOSmuggler:
> One thing I forgot to ask: what is the impact on users updating to a docker image featuring this new settings system? Is it something we can qualify/quantify?

yeah the whole point of this was to handle these settings in a manner non-destructive to the choices an instance's admin makes.  the previous method will override some choices.  this method respects choices and only sets unset settings to sane defaults.  as well as enforce envars, which was the previous behaviour.

there should be no other impact upon upgrade of or migration to this docker compose stack.  unless i've done something wrong.

i have some dev and prod instances using the previous system (not these changes) that i plan to test on before i will be happy with my changes.

UFOSmuggler:
More testing to go (particularly around the OIDC/other security/auth stuff) but so far:

**Create new docker compose misp-docker stack:**
ENABLE_DB_SETTINGS=true: Absolute minimum config settings and all CLI_only config directives created in config.php, remainder exist in system_settings table.
ENABLE_DB_SETTINGS=false: All config directives created in config.php.

**Change envar driven setting and restart or recreate container:**
ENABLE_DB_SETTINGS=true: Operator change reverted to envar in DB.  If previous value was stored in config.php, it contains the old value.
ENABLE_DB_SETTINGS=false: Operator change reverted to envar in config.php.

**Change safe default setting and restart or recreate container:**
ENABLE_DB_SETTINGS=true: Operator change persists in DB, config.php contains old value.
ENABLE_DB_SETTINGS=false: Operator change persists.

**Change CLI_only setting and restart or recreate container:**
ENABLE_DB_SETTINGS=true: Operator change persists, nothing created in database.
ENABLE_DB_SETTINGS=false: Operator change persists.

**Change bare minimum config file only setting and restart or recreate container:**
ENABLE_DB_SETTINGS=true: Operator change persists in database, config.php contains old value.
ENABLE_DB_SETTINGS=false: Operator change persists.

**Upgrade existing docker compose stack from prior to these changes:**
ENABLE_DB_SETTINGS=true: Existing settings kept, new settings set in database.  Old config.php structure maintained.
ENABLE_DB_SETTINGS=false: Existing settings kept, new settings set in config.php.  Old config.php structure maintained.

**Note**: Attempting to set MISP.redis_password to "" in the database results in it not being stored in DB at all. Its default value when unset is "" anyway. Recommend fixing by generating random strings for config passwords and secrets, to be stored in config.php.

Everything so far looks fine.

comparison of diagnostics pages (from vanilla as possible install):

old:
Screenshot from 2024-06-05 10-41-14

new:
Screenshot from 2024-06-05 11-41-26

UFOSmuggler:
Testing of configurations using AAD, OIDC, and LDAP resulted in identical configurations being made by the current misp-docker misp-core image, and one built with these changes.

In each case, all settings were generated correctly in the config.php file, and if system_settings table was being used, most of the config went in config.php with the remaining small number of config directives going in the databse.

UFOSmuggler (commented):
All review questions satisfactorily met.

ostefano:
I think we are ready to go.

I would add a line to the README file detailing which JSON file to modify when adding a new setting (something that we do on a regular basis, and hence we want to make as straightforward as possible).

Something like:

```
Adding a new setting and unsure where/how to edit the JSON files?

If it is just a default setting that is meant to be set if not already set by the user, add it here.
If it is a setting controlled by an ENV variable which is meant to override whatever is set, add it there.
Note that you can still specify a default value in the second case.
```

But I can add this myself in case.

Do you want me to merge, or further polish it?

ostefano (approved):

UFOSmuggler:
click the big shiny merge button please.  feel free to add text to documentation.

i might polish further at a later point in time.

PR #71, opened 2024-06-11
Update set_up_proxy(), add DEBUG envar, minor fixes

### LINKED ISSUE #57: Debug Configuration as an Environment Variable
There seems to be no way to instantiate a misp-core container with debug mode enabled.

While it is possible to enable debug mode through the UI, it may not always be usable. Consider the case where the container is broken and fails to authenticate despite showing the web UI. In such a case, it's tedious to login into the container to enable debug mode over the shell. It would be convenient if we could respin the container in debug mode amd look at the debug error logs shown on the web UI.

This can be fixed by adding a dedicated "MISP_DEBUG" variable in the .env file.

Obviously, enabling debug mode in a production environment is unsafe, so documentation should appropriately warn of the risk.

UFOSmuggler:
added missing PROXY_ENABLE envar to docker-compose.yml

add default value of false if PROXY_ENABLE unset

altered set_up_proxy() in configure_misp.sh to use json-based config system

fixed MISP.ca_path setting to use --force, since doesn't end in .pem

### DIFF
--- a/core/files/configure_misp.sh
+++ b/core/files/configure_misp.sh
@@ -13,6 +13,7 @@ source /utilities.sh
 [ -z "$OIDC_ENABLE" ] && OIDC_ENABLE="false"
 [ -z "$LDAP_ENABLE" ] && LDAP_ENABLE="false"
 [ -z "$ENABLE_DB_SETTINGS" ] && ENABLE_DB_SETTINGS="false"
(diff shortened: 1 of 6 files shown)


ostefano:
LGTM. Didn't know that Debug could accept int values 😑

ostefano (approved):

PR #75, opened 2024-06-12
Core - Enhanced the validity check of Nginx's dhparam.pem file during configuration phase

Aldenar:
Whenever I tried running the misp-core image, it'd get stuck on an infinite loop of Nginx trying to start up, then failing due to invalid DHParam.pem file.

For unknown reasons, the file existed, but was empty every time.

This PR fixes that by enhancing the dhparam.pem file check, adding a contents check of the file, and if anything other than a valid DH Params are found, regenerates the file.

### DIFF
--- a/core/files/entrypoint_nginx.sh
+++ b/core/files/entrypoint_nginx.sh
@@ -255,7 +255,7 @@ init_nginx() {
         echo "... TLS certificates found"
     fi
     
-    if [[ ! -f /etc/nginx/certs/dhparams.pem ]]; then
+    if [[ ! -f /etc/nginx/certs/dhparams.pem ]] || ! openssl dhparam -in /etc/nginx/certs/dhparam.pem -noout -check; then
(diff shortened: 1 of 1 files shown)


ostefano:
This looks to me like a workaround to a runtime (and singular) issue.
Not sure it would make sense to address it explicitly.

Aldenar:
On my server, I have tried resetting docker to a clean state several times, and always, the initial configuration got stuck on the dhparam file existing on disk, but being empty, thus leading to Nginx failing to start. 

This commit does not alter the overall flow of the configuration, only adds a content verification step that, in my opinion, should have been present from the start. 

I tried debugging why the file initially failed to generate, but that didn't lead to anything. This, however, worked fine. And finally allowed the container to start correctly.

ostefano:
Mmm, did you try with the latest docker version available from docker.com?

PR #77, opened 2024-06-18
Make SMTP host setting configurable

### LINKED ISSUE #78: Feature Request: Make SMTP host a configurable option
Right now email.php values are hard coded forcing user to use the mail subsystem in docker-compose.
While this is handy for local deployments, it's not suitable when you want to use the image in a different environment like k8s or even docker-compose deployment where another smtp service already exists.

damanzanofortra:
Allows to configure a SMTP server different from the one in docker-compose while making the docker-compose one the default.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -159,7 +159,7 @@ services:
       # Debug setting 
       - "DEBUG=${DEBUG}"
       # SMTP setting
-      - "SMTP_FQDN=${SMTP_FQDN:-mail}"
+      - "SMTP_FQDN=${SMTP_FQDN}"
   
   misp-modules:
     image: ghcr.io/misp/misp-docker/misp-modules:latest
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -158,6 +158,8 @@ services:
       - "MYSQL_DATABASE=${MYSQL_DATABASE:-misp}"
       # Debug setting 
       - "DEBUG=${DEBUG}"
+      # SMTP setting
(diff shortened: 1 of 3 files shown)


DaanWillems:
This is also causing some troubles for me, Nice fix!

template.env
```diff
@@ -44,6 +44,9 @@ BASE_URL=
 # store settings in db except those that must stay in config.php. true/false, defaults to false
 ENABLE_DB_SETTINGS=
 
+# Defines what mail system to use defaults to 'mail' which is provided in docker-compose.yml
+SMTP_FQDN=mail
```
ostefano, line 48: Since there is a default, this should be commented

docker-compose.yml
```diff
@@ -158,6 +158,8 @@ services:
       - "MYSQL_DATABASE=${MYSQL_DATABASE:-misp}"
       # Debug setting 
       - "DEBUG=${DEBUG}"
+      # SMTP setting
+      - "SMTP_FQDN=${SMTP_FQDN:-mail}"
```
ostefano: Since you default inside `entrypoint_nginx.sh` this should not have a default

ostefano (changes requested):

ostefano:
Left a few comments. Happy to merge once addressed.

damanzanofortra:
Addressed @ostefano comments in https://github.com/MISP/misp-docker/pull/77/commits/8f2a0fd2ff7deacfc07f64519c39c7d302ed9076

PR #87, opened 2024-07-04
fix: curl has to know about https_proxy

marjatech:
curl command will fail after a long timeout behind a corporate proxy.  
adding `-E` allows the user to influence this  by adding `https_proxy` environment variable to docker-compose file.


ostefano:
LGTM, thanks 🙏

PR #92, opened 2024-07-15
Update to Debian bookworm and python 3.12

ostefano:

PR #93, opened 2024-07-15
Replace Debian with Ubuntu 24.04 LTS

ostefano:


ostefano:
There is an underlying incompatibility with libcurl, tls, and `api.github.com` as used by composer.

Will resume investigation when 24.04.1 is released

ostefano:
Issue being investigated here https://github.com/curl/curl/issues/14154

ostefano:
The workaround disables SSL verification when downloading PHP dependencies when building the image for ARM64. Note that SSL validation is enabled for the AMD64 arch, and the packages downloaded are the same.

UFOSmuggler:
this looks fine as well but i only looked at x86_64 images.  same errors as noted in #92 

edit: actually i forgot to login and check the actual application, it is having errors.  investigating.

hmmm.  recreated and no more errors...  unable to recreate error... unsure what happened.

PR #113, opened 2024-08-15
Enable auditlog for misp-core & honouring image TAG variables for podman-compose pull 

joe-speedboat:
Hi
when using podman-compose up, it spamms the logs with:
[misp-core]    | sudo: unable to send audit message: Operation not permitted
This is due insufficient capabilities in podman, please add this line to allow audit log writes.
This can only write audit logs, and not configure/delete them...

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -43,6 +43,8 @@ services:
 
   misp-core:
     image: ghcr.io/misp/misp-docker/misp-core:latest
+    cap_add:
+      - CAP_AUDIT_WRITE
     build:
       context: core/.
       args:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -42,7 +42,7 @@ services:
       start_period: 30s
 
   misp-core:
-    image: ghcr.io/misp/misp-docker/misp-core:latest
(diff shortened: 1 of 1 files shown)


joe-speedboat:
Thanks @archham for pointing me to that!

joe-speedboat:
One more pull request.
latest tags in `docker-compose.yml` lead into different results between build and pull:
- build is honouring `*_TAG` 
- pull is always getting `latest` tag.
this is not reflecting ability to use <commit> docker image tags, but since `.env` can not reflect this logic, this is most you can do.
thanks Chris

joe-speedboat:
I created also an issue  #114

ostefano:
> Hi when using podman-compose up, it spamms the logs with: [misp-core] | sudo: unable to send audit message: Operation not permitted This is due insufficient capabilities in podman, please add this line to allow audit log writes. This can only write audit logs, and not configure/delete them...

Is this capability backward compatible? Aka, does it break docker compose? Can you test it?

ostefano:
Also, split the PR/issue please.

joe-speedboat:
I will split it into two requests

PR #117, opened 2024-08-15
added `*_TAG` vars instead of latest

### LINKED ISSUE #114: latest Tags in docker-compose.yml are hardcoded / Audit Log capabilities for misp-core
Hi

I found two minor misconfiguration in `docker-compose.yml`

- Missing Capabilities for misp-core container, which prevents writing auditlogs
- "latest" tags where hard code in docker image url, but should be tag variables from env

#113   Is the pull request for this two config changes

thanks for merging

joe-speedboat:
One more pull request.
latest tags in docker-compose.yml lead into different results between build and pull:

    build is honouring *_TAG
    pull is always getting latest tag.
    this is not reflecting ability to use docker image tags, but since .env can not reflect this logic, this is most you can do.
    thanks Chris

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -42,7 +42,7 @@ services:
       start_period: 30s
 
   misp-core:
-    image: ghcr.io/misp/misp-docker/misp-core:latest
+    image: ghcr.io/misp/misp-docker/misp-core:${CORE_TAG}
     build:
       context: core/.
       args:
@@ -167,7 +167,7 @@ services:
       - "SMTP_FQDN=${SMTP_FQDN}"
   
   misp-modules:
-    image: ghcr.io/misp/misp-docker/misp-modules:latest
+    image: ghcr.io/misp/misp-docker/misp-modules:${MODULES_TAG}
     build:
       context: modules/.
       args:


ostefano:
That is expected. Pull needs to fetch the latest available. Build would create a new latest, so when you bring up the container it would either use the pulled one (tagged latest) or the built one (tagged latest). Also note how you can build against a specific tag.

I am not planning to merge this.

joe-speedboat:
that breaks the meaning of the .env vars, sad but ok, i will move away, your thing

ostefano:
But those .env var are correctly named `Build time` variables, that is my point.

```
##
# Build-time variables
##

CORE_TAG=v2.4.195
MODULES_TAG=v2.4.195
PHP_VER=20190902
LIBFAUP_COMMIT=3a26d0a
```

joe-speedboat:
correct, and if you forex have installed:
CORE_TAG=v2.4.193
MODULES_TAG=v2.4.193
and pull images with podman-compose pull it will pull latest, which are now v2.4.195
then you result in env and docker-compose.yml working with images v2.4.193, but start v2.4.195, because of "latest"

this ended up really ugly in our env.

.env, docker-compose.yml and docker-images need to be consistent, always, ever.

or mention, that you never ever use podman-compose pull, since it breaks compability

thats my point

ostefano:
Alright, so you are saying:

1) build 2.4.193
2) happily use 2.4.193
3) pull latest
4) would expect to keep happily use 2.4.193, but 195 is executed

Correct? Would a new env var called "RUNNING_TAG" defaulting to "latest" but overridable as desired, fix the issue?

ostefano:
I see you like the solution, can you prepare a PR? Thx!

joe-speedboat:
RUNNING_TAG is even clearer.
but latest is always dangerous, since it can move to newer image than "misp-docker repo is reflecting", which cause problems.
its how i see, Im always to go save

ostefano:
To be honest, if you are deploying this in a DC, `docker-compose.yml` should be templated by something like `ansible` and the version of the image pinned somewhere, or at least this is how we do it and the spirit of this docker image.

joe-speedboat:
sure, we have dozens of apps like this, but if you add a single var to env or docker-compose.yml, 

we have to review and catch-up every single line change for this app. 

Its ... time consuming, so lets wait for the helm chart, it seems more promising than our discussion, thanks for your time!

If you look at awx forex, an operator upgrade takes us about 30s of work,  ... then automated QA ... then GO ... this is state of the art

but its opensource and all of us have limited ressources, thanks for all your time and effort @ostefano

ostefano (changes requested):
Create new variable RUNNING_TAG defaulting to latest

ostefano:
Closing because superseded by last build https://github.com/MISP/misp-docker/actions/runs/10538630557

PR #122, opened 2024-08-20
Extend customization of redis variables and unify defaults

ostefano:
@UFOSmuggler We had many issues recently due to some bugs in our defaults systems, and other most destructive issues caused by MISP rejecting settings when system_db_setting is enabled.

A new release of MISP is scheduled soon, but in the meanwhile it would be good to further tests these last changes.

### DIFF
--- a/core/files/configure_misp.sh
+++ b/core/files/configure_misp.sh
@@ -2,19 +2,8 @@
 
 source /rest_client.sh
 source /utilities.sh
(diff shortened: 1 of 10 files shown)


UFOSmuggler:
What's in the new MISP release?  Something related?

Some things cannot be moved into the system settings DB, and MUST be in the config.php file.  Is this what you are encountering?

PR #135, opened 2024-09-03
Run component updates in the background

oivindoh:
Currently, starting the image after bumping version numbers tends to be stuck for a very long time on component updates.

This PR runs these updates in the background rather than blocking startup waiting for the updates to be complete.

I don't really see a reason to wait for these to finish before proceeding with that we do after, and this change potentially reduces update downtime by a lot, depending on how time-consuming these processes are per update.

### DIFF
--- a/core/files/configure_misp.sh
+++ b/core/files/configure_misp.sh
@@ -353,7 +353,9 @@ init_settings() {
 update_components() {
     UPDATE_SUDO_CMD="sudo -u www-data"
     if [ ! -z "${DB_ALREADY_INITIALISED}" ]; then
-        UPDATE_SUDO_CMD="sudo -b -u www-data"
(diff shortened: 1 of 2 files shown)


ostefano:
Would it be possible to make it non-background the first time the image is run? (empty DB).

Reason is: enabling workflow modules will trigger a list of errors if this command is not run at least once.

Even better: make it optional (background option) via env var setting?

oivindoh:
@ostefano good catch!

I updated the PR to only run this in the background if the database was already initialised and then only if the BACKGROUND_COMPONENT_UPDATES env var is set

oivindoh:
@ostefano I think that takes care of all issues while avoiding giving users a foot cannon 👍

ostefano:
@oivindoh LGTM. Will merge as soon as I get some spare cycles to work on this 👍

ostefano:
Merged in other branch (required some refactory)

PR #136, opened 2024-09-03
Move settings util functions to utilities file to enable reuse

oivindoh:
These functions are very useful outside the context of configure_misp.sh.

I've just ported over my existing kludgy shell/python configuration hellscape to use these new functions sourced into customize_misp.sh


core/files/configure_misp.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
 
+MYSQLCMD="mysql -u $MYSQL_USER -p$MYSQL_PASSWORD -P $MYSQL_PORT -h $MYSQL_HOST -r -N $MYSQL_DATABASE"
```
ostefano: Do we need to export it?

oivindoh: Might as well export it early on, yeah

ostefano:
Merged in other branch (required some refactory)

PR #143, opened 2024-09-11
Add podman-systemd support

net-fall:
Added `container`, `volume` and `network` files for podman in **rootless** mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,71 @@ Custom root CA certificates can be mounted under `/usr/local/share/ca-certificat
       - "./rootca.pem:/usr/local/share/ca-certificates/rootca.crt"
 ```
 
+## Podman
+
+### Use Podman-systemd instead of Docker to:
+
+- Run containers in **rootless** mode
+- Manage containers with **systemd**
+- Write container descriptions in an **ignition** file and deploy them to an OS like **Fedora CoreOS** or similar (not covered in this documentation)
+
+### Copy the following directories and files:
+
+- Files from `podman-systemd` to `$USER/.config/containers/systemd/`
+- `template.vars` to `$USER/.config/containers/systemd/vars.env`
+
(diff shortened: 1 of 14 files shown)


ostefano:
Nice one!

I will add some more documentation marking this experimental (simply because of a really low bus factor), and then I'll merge.

Thanks again for this!

ostefano:
@urfin00djuce would it be possible to use `CORE_RUNNING_TAG` and `MODULES_RUNNING_TAG` (both defaulting to `latest`) to specify what images should be running?

ostefano (changes requested):
@urfin00djuce would it be possible to use `CORE_RUNNING_TAG` and `MODULES_RUNNING_TAG` (both defaulting to `latest`) to specify what images should be running?

net-fall:
@ostefano This might work, at least I can check it.
However, in this case, Podman won't be able to automatically update containers. As I mentioned in the `README.md` file, you can activate the `podman-auto-update.timer`. The container files already have the `AutoUpdate=registry` line in the `[Container]` section. This way, Podman will periodically check for a new version of the container. If a new version is available, Podman will pull it and restart the container. If the new version fails to start for some reason, Podman will try to start the previous version of the container. At least that's how it should work. Currently, the `CORE_TAG` and `MODULES_TAG` variables have a static number, so Podman will pull exactly this version of the container. And Podman will only be able to update the version when someone or something pulls a new fresh version of the **misp-docker** repository, modify `vars.env` file and restarts the containers. Or this process needs to be automated somehow.
Anyway, please let me know if you want to see `CORE_TAG` and `MODULES_TAG` specifically in the container files so I can verify this. It's also worth removing the section about `podman-auto-update.timer`, as it won't make sense in this case.

ostefano:
@urfin00djuce merged as part of another commit.

Thanks a lot for the contribution 👍

PR #146, opened 2024-09-16
Release MISP 2.5 and PHP 8.2

ostefano:


ostefano:
CC @iglocska (you can see how I build ssdeep)

PR #148, opened 2024-09-18
Add OIDC Custom Logout URL

### LINKED ISSUE #144: OIDC Custom Logout URL
Is it possible to add a custom logout url in the .env in which case OIDC_ENABLE=true adds this to the misp core configuration script?

```sh
sudo -u www-data /var/www/MISP/app/Console/cake Admin setSetting -q "Plugin.CustomAuth_custom_logout" https://id.misp-sandbox.org/realms/Strategic/protocol/openid-connect/logout?client_id=misp-corral&post_logout_redirect_uri=https://misp-sandbox.org/users/logout
```

image

Diogo-Rego:


core/files/configure_misp.sh
```diff
@@ -103,6 +103,9 @@ set_up_oidc() {
         }
     }" > /dev/null
 
+    # Set the custom logout URL for the OIDC plugin
```
ostefano: You want to do this only if the variable is defined

ostefano (changes requested):
Need to handle cases when the var is not defined

ostefano (approved):

ostefano (approved):

PR #153, opened 2024-09-20
fixes inability to configure $MYSQL_CMD variable

Amorik:
Makes it so that the $MYSQL_CMD variable can be set as a docker environmental variable to adjust the the command executed by the container to connect to a database. Usefully for making custom adjustments such as connecting via a socket instead of a network.


ostefano:
Can you explain a bit more the use case, and why you can not achieve the same thing by setting the MYSQL env vars individually?

Amorik:
The number of available environmental variables available to the client via the docker execution scripts is limited. For example, in my current setup I run mariadb on the local system (not docker) and bind mount the socket into the misp-core instance. This works with v2.4.194 (core)

compose file
```docker-compose.yml
...
volumes:
  - /var/run/mysqld/mysqld.sock:/var/run/mysqld/mysqld.sock
environment:
  - "MYSQL_SOCKET=${MYSQL_SOCKET}"
  - "MYSQLCMD=${MYSQLCMD}"
...
```

.env file
```.env
...
MYSQL_SOCKET=/run/mysqld/mysqld.sock
MYSQLCMD="mysql -u $MYSQL_USER -p$MYSQL_PASSWORD --socket=$MYSQL_SOCKET -r -N $MYSQL_DATABASE"
...
```

While most users will primarily use a docker container for mariadb/mysql as indicated in the setup documentation sometimes more advanced setups use different components in different places. Another example of this would be using a remote database where one may one to utilize SSL/TLS encryption, or network compression; options available via the client and sometimes required by the server.

While these could be added to the docker misp setup configs it may be easier (short term anyways) to simply allow advanced users to modify the client command instead. Though its not documented currently.

ostefano:
My point is also that the variable will not populate the database.php file used by MISP, so I am wondering whether a better handling of the many MYSQL variables would be a better solution instead.

Amorik:
My apologies, I did not even consider the PHP database config; i was only looking at the client command - I over simplified the issue.

Considering that the cake PHP and the client configurations dont seem to align 100% in terms of options I agree a different solution would be better. Acceptable variables would likely also have to be limited to what Cake PHP supports.

I'm not proficient in PHP but will look at how to address the issue differently.

PR #161, opened 2024-10-02
Fix OIDC_ENABLE Toggle Logic in MISP Environment Configuration

### LINKED ISSUE #158: MISP Env `OIDC_ENABLE=true`
MISP Env `OIDC_ENABLE=true` not working correctly, where when doing `OIDC_ENABLE=true` and then `OIDC_ENABLE=false` it still remains with `OIDC_ENABLE=true`

Diogo-Rego:


core/files/configure_misp.sh
```diff
@@ -72,46 +72,72 @@ GPGEOF
...
+                \"roles_property\": \"\",
+                \"role_mapper\": \"\",
+                \"default_org\": \"\"
+            }
+        }" > /dev/null
 
-    # Disable password confirmation as stated at https://github.com/MISP/MISP/issues/8116
-    sudo -u www-data /var/www/MISP/app/Console/cake Admin setSetting -q "Security.require_password_confirmation" false
+        # Use sed to remove the OidcAuth.Oidc entry from the 'auth' array in the config.php
+        sudo sed -i "/'auth' =>/,/)/ { /0 => 'OidcAuth.Oidc',/d; }" /var/www/MISP/app/Config/config.php
```
ostefano: Use `sudo -u www-data` here

ostefano (changes requested):
What if I have

core/files/configure_misp.sh
Diogo-Rego: @ostefano  like this?
```
# Use sed to remove the OidcAuth.Oidc entry from the 'auth' array in the config.php
sudo -u www-data sed -i "/'auth' =>/,/)/ { /0 => 'OidcAuth.Oidc',/d; }" /var/www/MISP/app/Config/config.php

# Remove the custom logout URL
sudo -u www-data sed -i "/'CustomAuth_custom_logout' =>/d" /var/www/MISP/app/Config/config.php
```

ostefano: Yes, all commands that edit MISP state and file should use the same user/permission used by MISP

ostefano (approved):

PR #164, opened 2024-10-04
Add additional PHP settings and cookie configurations

Diogo-Rego:


Diogo-Rego:
@ostefano  confirm cookie configuration in PHP

ostefano (changes requested):
Can you prefix the PHP vars with PHP_SESSION and add some more documentation for each option?

Diogo-Rego:
> Can you prefix the PHP vars with PHP_SESSION and add some more documentation for each option?

@ostefano in *README.md* or *template.env*

ostefano:
> > Can you prefix the PHP vars with PHP_SESSION and add some more documentation for each option?
> 
> @ostefano in _README.md_ or _template.env_

Inside `template.env`

Diogo-Rego:
@ostefano 
```dotenv
# PHP FPM Configuration

## Basic PHP Settings
# Sets the maximum amount of memory a PHP script can consume.
# PHP_MEMORY_LIMIT=2048M
# Specifies the maximum time in seconds a PHP script can take to execute before being terminated.
# PHP_MAX_EXECUTION_TIME=300
# Defines the maximum size allowed for file uploads in a PHP script.
# PHP_UPLOAD_MAX_FILESIZE=50M
# Specifies the maximum size allowed for POST data that can be sent to PHP.
# PHP_POST_MAX_SIZE=50M
# Defines the maximum time in seconds that PHP should spend parsing input data, such as $_GET, $_POST, and $_COOKIE.
# PHP_MAX_INPUT_TIME=300

## Additional PHP Settings
# Sets the timeout (in minutes) of inactivity after which the user session expires.
# PHP_TIMEOUT=60
# Specifies the time (in minutes) that a session cookie should remain valid in the client's browser.
# PHP_COOKIE_TIMEOUT=10080
# Sets the default PHP configurations, which can be adjusted according to the application’s needs.
# PHP_DEFAULTS=php
# When enabled, this parameter forces the automatic regeneration of the session ID on each request.
# PHP_AUTO_REGENERATE=false
# When enabled, it checks the user agent on each request to help prevent session hijacking attacks.
# PHP_CHECK_AGENT=false
# When set to true, session cookies will only be sent over secure (HTTPS) connections.
# PHP_COOKIE_SECURE=true
# Defines the domain where the session cookie is valid. It should be left empty if you want the cookie to be valid only on the current domain.
# PHP_COOKIE_DOMAIN=
# Defines the SameSite policy for cookies, which controls how cookies are sent with requests from other sites. The "Lax" value allows cookies to be sent with top-level navigation requests and in the case of GET requests.
# PHP_COOKIE_SAMESITE=Lax
```

**OR**

```dotenv
# PHP FPM Configuration

## Basic PHP Settings
# Maximum memory a PHP script can use.
# PHP_MEMORY_LIMIT=2048M
# Maximum execution time for a PHP script in seconds.
# PHP_MAX_EXECUTION_TIME=300
# Maximum file upload size for PHP scripts.
# PHP_UPLOAD_MAX_FILESIZE=50M
# Maximum size for POST data sent to PHP.
# PHP_POST_MAX_SIZE=50M
# Maximum time PHP spends parsing input data in seconds.
# PHP_MAX_INPUT_TIME=300

## Additional PHP Settings
# Timeout (in minutes) for user session inactivity.
# PHP_TIMEOUT=60
# Session cookie validity period in minutes.
# PHP_COOKIE_TIMEOUT=10080
# Default PHP configurations.
# PHP_DEFAULTS=php
# Automatically regenerate session ID on each request.
# PHP_AUTO_REGENERATE=false
# Check user agent on each request for security.
# PHP_CHECK_AGENT=false
# Only send session cookies over HTTPS.
# PHP_COOKIE_SECURE=true
# Domain for session cookie validity (leave empty for current domain).
# PHP_COOKIE_DOMAIN=
# SameSite policy for cookies ("Lax" allows top-level navigation).
# PHP_COOKIE_SAMESITE=Lax
```

ostefano:
Both work (personally like the second one as it is more concise).

Would it make sense to prefix some of the vars related to session management with `PHP_SESSION`?

Diogo-Rego:
> Both work (personally like the second one as it is more concise).
> 
> Would it make sense to prefix some of the vars related to session management with `PHP_SESSION`?

@ostefano 
In my opinion yes, it makes sense to prefix session-related variables with PHP_SESSION. This improves clarity and distinguishes session-specific settings from general PHP configurations, especially when managing various environment variables.MISP also puts these like this by default. What do you think ?

Diogo-Rego:
@ostefano confirm?

ostefano:
Sounds good! 👍

core/files/configure_misp.sh
```diff
@@ -219,6 +219,35 @@ set_up_aad() {
...
+                \"session.cookie_domain\": \"${PHP_COOKIE_DOMAIN}\",
+                \"session.cookie_samesite\": \"${PHP_COOKIE_SAMESITE}\"
+            }
+        }
+    }" > /dev/null
+
+    echo "... Session configured"
+}
+
+set_up_proxy() {
```
ostefano, line 242: I guess this is part of a future PR?

template.env
```diff
@@ -178,13 +178,38 @@ SYNCSERVERS_1_PULL_RULES=
...
 # PHP_MAX_EXECUTION_TIME=300
+# Maximum file upload size for PHP scripts.
 # PHP_UPLOAD_MAX_FILESIZE=50M
+# Maximum size for POST data sent to PHP.
 # PHP_POST_MAX_SIZE=50M
+# Maximum time PHP spends parsing input data in seconds.
 # PHP_MAX_INPUT_TIME=300
 
+## Additional PHP Settings
+# Timeout (in minutes) for user session inactivity.
```
ostefano: Please change variable names as discussed

ostefano (changes requested):
See comments for changes requested

Diogo-Rego:
@ostefano can you confirm

ostefano (approved):

ostefano:
@Diogo-Rego tried the PR and first execution prints:

```
misp-core-1     | MISP | Set Up Session ...
misp-core-1     | Provided new config is not array, `NULL` given.
misp-core-1     | ... Session configured
```

Is that expected?

ostefano (commented):
To investigate error message

Diogo-Rego:
> misp-core-1     | Provided new config is not array, `NULL` given.

I'll confirm the error

Diogo-Rego:
@ostefano confirm

Diogo-Rego:
@ostefano Can you confirm if it works?

ostefano:
@Diogo-Rego Didn't have spare cycles today yet. Will get to it as soon as I can.

ostefano:
Works, merging it.

PR #180, opened 2024-10-31
Fix OIDC SCOPES

Diogo-Rego:
This modification allows the `scopes` field in the OIDC configuration to be added only if an `OIDC_SCOPES` variable is provided. For providers like Keycloak, scopes may not be mandatory, as permissions and access can be managed in other ways. Therefore, the configuration has been adjusted to make the `scopes` field optional, adding it only when needed.


ostefano:
LGTM

ostefano (approved):

PR #184, opened 2024-11-12
$OIDC_SCOPES must be an array, should not be in quotes

capile:
OIDC_SCOPES (OidcAuth.scopes) should not be quoted before setting to config, as it expects a JSON array. As there's `jq` available at the container, we can properly test if the variable is indeed an array before pushing to config

### DIFF
--- a/core/files/configure_misp.sh
+++ b/core/files/configure_misp.sh
@@ -101,13 +101,13 @@ set_up_oidc() {
             }
         }" > /dev/null
 
-        # Check if OIDC_SCOPES is set and not empty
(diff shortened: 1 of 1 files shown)


ostefano:
@JordanBarnartt CC

JordanBarnartt:
Makes sense to me.  Looking back, my original PR (https://github.com/MISP/misp-docker/pull/170/files) did not include quotes, but it also made the field mandatory, and the quotes were added as part of another PR (https://github.com/MISP/misp-docker/pull/180/files).

ostefano:
@Diogo-Rego can you give it a look?

ostefano (approved):

PR #186, opened 2024-11-14
temporary response should not enable a successful healthcheck or cache

### LINKED ISSUE #185: Cache-Control and RollingUpdate strategy
There is no cache-control header neither on startup nginx or the final running container. This affects any server running behind a CDN or reverse proxy, as even sensitive pages like the `/users/login` lack a cache control definition. This causes OidcAuth to fail as the session is not properly set.

Also the startup instance returns true on health checks, to a `RollingUpdate` strategy would terminate the running container before the new one is ready to take over.

We propose to mitigate these issues that:

1. Startup nginx is disbled or fails healthchecks. If it is kept, definitely add proper cache-control headers maybe even retry-after ones (so API calls aren't lost). By Startup nginx we mean the "MISP is loading..." phase (is it required for any process?)

2. Add cache-control headers to any .php processing result. Or somehow manage to put those on every response on PHP code (in case some endpoints do have it implemented). They could be implemented on nginx whenever they are not set by PHP code.

capile:
the temporary response sent while additional configuration tasks are running preparing the container should not enable a successful healthcheck (thus the 503 response) or caching in upstream proxies (then the cache-control)

fixes: #185

### DIFF
--- a/core/files/var/www/html/index.php
+++ b/core/files/var/www/html/index.php
@@ -1,3 +1,10 @@
+<?php
+$proto = (isset($_SERVER['SERVER_PROTOCOL']))?($_SERVER['SERVER_PROTOCOL']):('HTTP/1.1');
+header($proto.' 503 Service Unavailable', true);
+header('cache-control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0');
+header('retry-after: 30');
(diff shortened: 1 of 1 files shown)


ostefano:
LGTM

Issue #188, opened 2024-11-27
Slow startup and performance on NFS or slow volume mounts


capile:
Using NFS (AWS EFS), S3 or other slow performance permanent storage can slow the startup time significantly (due to the time it takes for rsync + find + chmod + chown on these disks), and potentially affect performance of any module that uses the contents of `/var/www/MISP/app/files`.

However, these volume mounts are the only that allow the `ReadWriteMany` access mode and therefore a `RollingUpdate` deployment strategy (the alternative is to terminate and recreate, having a downtime whenever the service restarts).

PR #189, opened 2024-11-27
FIX: Slow startup and performance on NFS or slow volume mounts #188

capile:
Addresses issue #188 

Adds a bit of documentation on how to enable a different `ATTACHMENTS_DIR` and what other folders need to be mounted instead of the `/var/www/MISP/app/files`.

Places the attachments folder outside of the `files/` structure, that will remain basically the same as the distribution, with the exception of `files/img/`and `files/terms/`, folders that should be affected by a NFS filesystem.

During initialization, checks if the `files/` folder was mounted before performing any tasks on it.


core/files/entrypoint_nginx.sh
```diff
@@ -150,21 +159,29 @@ update_misp_data_files(){
...
-    echo "... chmod -R 0550 files /var/www/MISP/app/files" && find /var/www/MISP/app/files -not -perm 550 -type f -exec chmod 0550 {} +
-    # Directories are also writable, because there seems to be a requirement to add new files every once in a while
-    echo "... chmod -R 0770 directories /var/www/MISP/app/files" && find /var/www/MISP/app/files -not -perm 770 -type d -exec chmod 0770 {} +
-    # We make 'files' and 'tmp' (logs) directories and files user and group writable (we removed the SGID bit)
-    echo "... chmod -R u+w,g+w /var/www/MISP/app/files" && chmod -R u+w,g+w /var/www/MISP/app/files
+    # If $MISP_APP_FILES_PATH was not changed since the build, skip file updates there
+    FILES_VERSION=
+    MISP_APP_FILES_PATH=/var/www/MISP/app/files
+    if [ -f ${MISP_APP_FILES_PATH}/VERSION ]; then
+        FILES_VERSION=$(cat ${MISP_APP_FILES_PATH}/VERSION)
```
ostefano, line 166: Can you add a line echoing the parsed version?

ostefano (changes requested):
As per comment

core/files/entrypoint_nginx.sh
capile: Ola Stefano, sure, but I messed up and used the `master` branch instead of an issue, and now it's all messed up -- I'm re-creating this PR with the echo and the right branch

Please ignore the other updates on this PR

PR #192, opened 2024-12-02
FIX: Slow startup and performance on NFS or slow volume mounts #188

capile:
Replaces PR #189 

Adds a bit of documentation on how to enable a different ATTACHMENTS_DIR and what other folders need to be mounted instead of the /var/www/MISP/app/files.

Places the attachments folder outside of the files/ structure, that will remain basically the same as the distribution, with the exception of files/img/and files/terms/, folders that should be affected by a NFS filesystem.

During initialization, checks if the files/ folder was mounted before performing any tasks on it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -114,7 +114,6 @@ services:
       - "ADMIN_ORG=${ADMIN_ORG}"
       - "ADMIN_ORG_UUID=${ADMIN_ORG_UUID}"
       - "GPG_PASSPHRASE=${GPG_PASSPHRASE}"
-      - "ATTACHMENTS_DIR=${ATTACHMENTS_DIR:-/var/www/MISP/app/files}"
       # OIDC authentication settings
       - "OIDC_ENABLE=${OIDC_ENABLE}"
       - "OIDC_PROVIDER_URL=${OIDC_PROVIDER_URL}"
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -114,6 +114,7 @@ services:
       - "ADMIN_ORG=${ADMIN_ORG}"
       - "ADMIN_ORG_UUID=${ADMIN_ORG_UUID}"
       - "GPG_PASSPHRASE=${GPG_PASSPHRASE}"
+      - "ATTACHMENTS_DIR=${ATTACHMENTS_DIR:-/var/www/MISP/app/files}"
```
ostefano: defaults for env vars should be defined here, but inside the json file

template.env
```diff
@@ -53,6 +53,8 @@ ENABLE_DB_SETTINGS=
 ENCRYPTION_KEY=
 # enable background updates. defaults to false
 ENABLE_BACKGROUND_UPDATES=
+# use a different attachments_dir. configure this folder to use persistent storage
+ATTACHMENTS_DIR=/var/www/MISP/app/files
```
ostefano: Default should be unset

ostefano (changes requested):

capile:
changes committed, but as it is, unsetting the `ATTACHMENTS_DIR` after if was set does not revert it back to `/var/www/MISP/app/files` (should it?)

If it should, I suggest to add it to the `/entrypoint.sh`

ostefano:
This has been merged as a squash commit here https://github.com/MISP/misp-docker/commit/6599e572970a9fb20843b386d2a5f7f0af22a79e

Closing this.

PR #202, opened 2025-01-08
Include rsyslog for local syslogger

cudeso:
Possible rsyslog file can be: /etc/rsyslog.d/40-misp.conf

```
global(parser.PermitSlashInProgramname="on")

if (
    (
        $programname == '/var/www/MISP/app/tmp/logs/' or
        $programname == 'MISP-SYNC' or
        $programname == 'MISP-SCRIPTS'
    ) and not (
        $msg contains 'contracts.canonical.' or
        $msg contains 'ubuntupro.http'
    )
)

then /var/www/MISP/app/tmp/logs/mispsyslog.log

& stop
```

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -262,6 +262,9 @@ EOF
         touch /var/www/MISP/.git/ORIG_HEAD && chmod 0600 /var/www/MISP/.git/ORIG_HEAD && chown www-data:www-data /var/www/MISP/.git/ORIG_HEAD
 EOF
 
+    # Include rsyslog to support syslogger
+    RUN apt-get update ; apt-get install -y rsyslog
+
     # Copy all our image specific files to appropriate locations
     COPY files/ /
     ENTRYPOINT [ "/entrypoint.sh" ]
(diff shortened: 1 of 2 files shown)


core/Dockerfile
```diff
@@ -262,6 +262,9 @@ EOF
         touch /var/www/MISP/.git/ORIG_HEAD && chmod 0600 /var/www/MISP/.git/ORIG_HEAD && chown www-data:www-data /var/www/MISP/.git/ORIG_HEAD
 EOF
 
+    # Include rsyslog to support syslogger
+    RUN apt-get update ; apt-get install -y rsyslog
```
ostefano: This needs to be moved where the layer is built

cudeso: Do you mean as part of the `apt-get install` at https://github.com/MISP/misp-docker/blob/master/core/Dockerfile#L169   ?

ostefano: Correct

cudeso: Good, should be fixed now!

ostefano (approved):

ostefano:
Merging but `/etc/rsyslog.d/40-misp.conf` will not be part of this PR.

PR #203, opened 2025-01-08
Make it easier to run in Kubernetes

### LINKED ISSUE #112: Helm Chart
Hello,

Would a helm chart be a feature that the MISP project is interested in?  If so, would this repo be the appropriate place for that feature or would a different or even new MISP Helm repo be more appropriate?

Helm does offer a good deployment method for running MISP on production workloads.  The main project has some discussion about personal Helm charts, but nothing about an officially supported option.

jhuntwork:
These changes make running in Kubernetes simpler. The main goal here is to split out nginx, php-fpm and crond as separate processes that execute as PID 1 in their own container within a pod. This lets Kubernetes itself supervise those processes instead of the intermediary supervisord. For example, it allows Kubernetes to restart nginx or php-fpm individually in a container if that becomes necessary.

We tried to make these changes as unobtrusive as possible, so that it will be easier to test and doesn't interrupt existing functionality. We did this by testing for the presence of the `KUBERNETES_SERVICE_HOST` variable which will be present if this container is running in that environment. We then added new k8s specific entrypoint files that source and generally follow the logic of the existing entrypoint files, with the exception two major rules:
- Let the php-fpm container configure the application specific settings, since that is coupled tightly with php
- Don't try to manage any current running processes (like sending reload signals). Instead use 'exec' at the end so that the respective service becomes PID 1, and rely on eventual consistency to grant that the whole pod will reach a ready state once all configuration is done and all services are running.

This is also designed to use the same image for all three containers, but just modify their entrypoint script slightly for each case.

There are still a few outstanding things that should probably be improved:

1. It's unclear if the nginx container and php-fpm both need to write to /var/www/MISP (the app store). If they do, then this should be a shared volume between the two pods, which is easy enough. In my own implementation I build a custom image that moves /var/www/MISP to /MISP at build time and then at run time (after the shared mount is created) rsync /MISP to /var/www/MISP. There is probably a better solution but this is what we currently have.
2. Getting _all_ logging to output to stdout which is standard for containers, especially in kubernetes. This lets us do things like send to a full-featured logging service and do filtering/routing there.
3. Re-working the background workers into something different. I understand this is an upstream feature change, but the current implementation _expects_ that the workers will be running on the same host as the main php process and are discoverable via PID which does not hold true if they are containerized/scalable elsewhere.

Looking forward to feedback and happy to adjust where ever it makes sense.


ostefano:
Thank you for this PR @jhuntwork , this is awesome stuff 🎄

I will need a bit of time due to day-job constraints, but will get to it eventually.

Feel free to reach out to me on Gitter, so we can iterate quickly on this.

ostefano:
@oivindoh can you give it a look? Looks reasonable to be, but I would like this not to clash with your current way of doing things on K8S. Likewise, if you have suggestions, please add them here.

oivindoh:
Sorry, it seems I never actually hit send on the comment I wrote here earlier...

I think this looks good and is a sensible change. Perhaps we could default to standard behaviour if the container name is not php,nginx or cron - otherwise we'll need to surface some kind of breaking change note to end-users.

I'll give this a whirl in my test env before the end of the week.

core/files/entrypoint.sh
```diff
@@ -68,5 +68,22 @@ export PHP_SESSION_COOKIE_SAMESITE=${PHP_SESSION_COOKIE_SAMESITE:-Lax}
 export NGINX_X_FORWARDED_FOR=${NGINX_X_FORWARDED_FOR:-false}
 export NGINX_SET_REAL_IP_FROM=${NGINX_SET_REAL_IP_FROM}
 
-# start supervisord using the main configuration file so we have a socket interface
-/usr/bin/supervisord -c /etc/supervisor/supervisord.conf
+if [ -n "$KUBERNETES_SERVICE_HOST" ]; then
+  case "$CONTAINER_NAME" in
```
oivindoh, line 72: Really a shame the container name isn't exposed to the container environment by default.

core/files/entrypoint_cron.sh
```diff
@@ -30,6 +30,11 @@ if [[ ! -p /tmp/cronlog ]]; then
     mkfifo -m 777 /tmp/cronlog
 fi
 
+if [ -n "$KUBERNETES_SERVICE_HOST" ]; then
+    tail -f /tmp/cronlog &
+    exec cron -l -f
```
oivindoh, line 35: We're missing some setup for the default cron jobs to be able to run here (www-data /var/www/MISP/app/Console/cake Admin updateGalaxies and friends). Not sure it's feasible to run these in their own container.

Maybe the best solution is to rewrite them all to run API calls / pymisp against localhost with supplied admin_key?

core/files/entrypoint_fpm.sh
```diff
@@ -28,6 +28,7 @@ change_php_vars() {
         sed -i "s|.*session.save_path = .*|session.save_path = '$(echo $REDIS_HOST | grep -E '^\w+://' || echo tcp://$REDIS_HOST):$REDIS_PORT?auth=${ESCAPED}'|" "$FILE"
         sed -i "s/session.sid_length = .*/session.sid_length = 64/" "$FILE"
         sed -i "s/session.use_strict_mode = .*/session.use_strict_mode = 1/" "$FILE"
+        sed -i "s|session.cookie_domain = .*|session.cookie_domain = ${BASE_URL}|" "$FILE"
```
oivindoh, line 31: ❤️

oivindoh:
This seems to work well (with the exception of default cron jobs failing to run), with the three containers set up and the CONTAINER_NAME var supplied to all, as well as a shared emptyDir volume for /var/run/php between nginx and php containers.

I added a readinessprobe to count the number of socket files present there for both the nginx and php containers in my test deployment.

```yaml
          readinessProbe:
            exec:
              command:
                - /bin/sh
                - -c
                - 'test $(ls /run/php/php*-fpm*sock | wc -l) -eq 1'
```

Would be really nice to get all the PHP logs redirected to stdout. There's CakeLog in bootstrap.php responsible for app/tmp/logs/[error|debug].log, then here are also other hardcoded log files like ProcessTool (exec-errors.log), ServerSyncTool (server-sync.log), AppExceptionRenderer (fatal_error.log). The hardcoded ones certainly require code changes in MISP upstream.

I expected to be able to at least change boostrap.php to write to php://stdout, but so far no luck on that front here.

core/files/entrypoint_cron.sh
oivindoh: On second thought, a better solution would befor us to supply an example set of CronJob manifests to deploy alongside MISP.

core/files/entrypoint.sh
```diff
@@ -68,5 +68,22 @@ export PHP_SESSION_COOKIE_SAMESITE=${PHP_SESSION_COOKIE_SAMESITE:-Lax}
 export NGINX_X_FORWARDED_FOR=${NGINX_X_FORWARDED_FOR:-false}
 export NGINX_SET_REAL_IP_FROM=${NGINX_SET_REAL_IP_FROM}
 
-# start supervisord using the main configuration file so we have a socket interface
-/usr/bin/supervisord -c /etc/supervisor/supervisord.conf
+if [ -n "$KUBERNETES_SERVICE_HOST" ]; then
```
oivindoh, line 71: I'd change this to `if [ -n "$KUBERNETES_SERVICE_HOST" ] && [ -n "$CONTAINER_NAME" ]; then` to ensure we just do the default thing if CONTAINER_NAME is missing. This ensures existing setups aren't immediately broken by updating.

core/files/entrypoint_cron.sh
jhuntwork: Ah, ok, thanks. Admittedly I didn't look too far into the cron jobs.

jeremiah-RENISAC: @oivindoh I took a stab at setting this up: https://github.com/jeremiah-RENISAC/misp-docker/tree/kubernetes_config/core/files/k8s_cronjobs

jeremiah-RENISAC:
I'd love to see this picked up again - I took a stab at creating Kubernetes cronjob configurations for the standard jobs (except for updating Object Templates, which has no equivalent API call), adding the container_name check to prevent breaking existing setups, and resolving the merge conflicts that have cropped up in the last half year.

Anything else needed to get this going again? And I'm still a bit new to GitHub - should I submit my own PR from my branch, or is there a more elegant way to leverage this existing one?

https://github.com/jeremiah-RENISAC/misp-docker/tree/kubernetes_config

ostefano:
Since we merged https://github.com/MISP/misp-docker/pulls?q=is%3Apr+is%3Aclosed I will be closing this

PR #204, opened 2025-01-09
Remove doubled X-XSS-Protection Header

marjatech:
since it gets already set by MISP itself. [Code](https://github.com/MISP/MISP/blob/ec91464441443e2043e5616489ac25a4ab4575b6/app/Controller/AppController.php#L160)

[OWASPs](https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html#recommendation_1) recommendation for this header is:
> Do not set this header or explicitly turn it off. `X-XSS-Protection: 0`

image


ostefano:
@marjatech this is added by MISP only for REST calls though, right?

ostefano (commented):
See my other comment

marjatech:
I think it's added only for **non**-REST calls, which should be fine. 
And as i wrote OWASPs recommendation is anyway not to set it at all or set it to 0.

PR #211, opened 2025-01-15
Avoid removing terms and conditions files

cudeso:
Update core/files/entrypoint_nginx.sh


ostefano:
Why is this needed? Can `terms` be updated by MISP users and not just by MISP developers?

HugeekMcGill:
I also have issue (Didn't dig deep yet) where the term files is overwritten after every rebuild/upgrade.

cudeso:
> Why is this needed? Can `terms` be updated by MISP users and not just by MISP developers?

When you install MISP you can add your custom terms and conditions in `MISP/app/files/terms` (can be an HTML file). You can then enable the new terms via the `MISP.terms_file` setting.

cudeso:
> I also have issue (Didn't dig deep yet) where the term files is overwritten after every rebuild/upgrade.

The `--delete` in the rsync (of core/files/entrypoint_nginx.sh) removes your term file.

ostefano (approved):

ostefano:
Merged.

@cudeso on a separate note, could you document in the README the example of a rsyslog configuration?
Or add a sane default so it is fully enabled.

Referring to:
```
global(parser.PermitSlashInProgramname="on")

if (
    (
        $programname == '/var/www/MISP/app/tmp/logs/' or
        $programname == 'MISP-SYNC' or
        $programname == 'MISP-SCRIPTS'
    ) and not (
        $msg contains 'contracts.canonical.' or
        $msg contains 'ubuntupro.http'
    )
)

then /var/www/MISP/app/tmp/logs/mispsyslog.log

& stop
```

PR #231, opened 2025-03-12
Update docker-compose.yml

dying-breed:
Removed version attribute from docker-compose.yml: To address the warning WARN[0000] misp-docker/docker-compose.yml: the attribute 'version' is obsolete, it will be ignored, please remove it to avoid potential confusion.

Eliminate unnecessary warnings.


ostefano:
Last commit removed it already (see master)

PR #239, opened 2025-04-08
make PHP timezone setting configurable

marjatech:
since this is where MISP takes its times from


ostefano (changes requested):
You need to set a default in entrypoint.sh

ostefano (approved):

PR #247, opened 2025-04-24
add: [oidc] support mixedAuth setting

righel:
add support for new `OidcAuth.mixedAuth` setting
related to: https://github.com/MISP/MISP/pull/10329


core/files/configure_misp.sh
```diff
@@ -97,7 +97,8 @@ set_up_oidc() {
                 \"client_secret\": \"${OIDC_CLIENT_SECRET}\",
                 \"roles_property\": \"${OIDC_ROLES_PROPERTY}\",
                 \"role_mapper\": ${OIDC_ROLES_MAPPING},
-                \"default_org\": \"${OIDC_DEFAULT_ORG}\"
+                \"default_org\": \"${OIDC_DEFAULT_ORG}\",
+                \"mixedAuth\": \"${OIDC_MIXEDAUTH}\"
```
ostefano: Should this be added only when used?

ostefano (commented):
See comments

core/files/configure_misp.sh
righel: It defaults to false (previous) behavior, so indeed could be added only when needed like the `issuer`.

righel: changed by 3c8a63077b44523cbbc912b6a02617719a513546. should I remove it from template.env too? if so, how do users know of it's existence?

ostefano (changes requested):
Sorry, dropped the ball a bit on this. Review comments below

What you want to do is 
- leave in template.env a possible value (not necessarily default) commented
- in entrypoint.sh set the variable to a sane default (false)
- inside configure_misp.sh keep it a simple assignment

ostefano (approved):

PR #249, opened 2025-04-29
add modification crontab

### LINKED ISSUE #248: how can I modify crontab
Hello,

Quick question, how can I modify the crontab of "pullAll" (here https://github.com/MISP/misp-docker/blob/44e417fb5d2259a824ffd48e60df1e9316959d0d/core/files/entrypoint_cron.sh#L17C58-L17C65 ).

If I want to have a pullall, every 10 min, where should I add/modify it? docker-compose?

arteta22000:
Hello,

it was discussed here: https://github.com/MISP/misp-docker/issues/248
This PR adds the ability to modify pull and push crontab.


ostefano (changes requested):
Defaults should be set inside the main entry point script, so if the env does not have those vars set, a default is ensured.

Also, template should have those vars commmented.

arteta22000:
> Defaults should be set inside the main entry point script, so if the env does not have those vars set, a default is ensured.
> 
> Also, template should have those vars commmented.

Hello,

ok, I added this, ok for you?

ostefano (changes requested):
Almost. The default should be set inside the main entrypoint (entrypoint.sh) and not inside the cron one.

arteta22000:
> Almost. The default should be set inside the main entrypoint (entrypoint.sh) and not inside the cron one.
good ?

ostefano (approved):

ostefano (approved):

PR #255, opened 2025-05-17
Updating entrypoint_fpm.sh to fix the sed command for setting timezone.

darshannnn:
I have been trying to update the timezone on my MISP instance to, 'Australia/Sydney' but hasn't been working.
Seems like the sed command for setting the timezone isn't working correctly.
I hand to add the -E flag and use, '|' instead of '/' in the sed command as timezone has a, '/' as well.
I tested this and this has correctly set `date.timezone` in `/etc/php/*/fpm/php.ini`


ostefano:
Is '-E' mandatory? That is usually not required to just use '|' instead of '/'.

darshannnn:
'-E' seems to be necessary for the regex to work to do the `;?`  match like in lines 41, 43 and more in the same file that updates www.conf.
I've updated that line to be more consistent with other sed commands in the file.
The previous sed was checking for all the possible whitespaces.

PR #270, opened 2025-07-11
Make client_secret optional for OIDC + Add code_challenge_method for PKCE as configurable env var

chaguser:
This PR proposes two improvements to the OIDC configuration in the 'configure_misp.sh' script used in the misp_core container.

**What this changes:**

1. **Make client_secret optional:**

     * The current script requires the OIDC_CLIENT_SECRET env variable to be set, even when it's not needed.
     * In modern setups using **PKCE**, a client_secret is often not used (e.g, public clients).
     * Forcing it causes the OIDC PHP Library (jakub-onderka/openid-connect-php) to assume a confidential client, which unintentionally triggers **JWT Secured Authorization Requests (JAR)**  - not something MISP explicitly enables, but a side-effect of the library's behavior,  that isn't expected by default.

2. **Adds support for configuring code_challenge method via environment variable:** 

     * Introduces OIDC_CODE_CHALLENGE_METHOD in template.env
     * Automatically writes it to the config.php in the MISP container via configure_misp.sh
     * This gives organizations using PKCE more control whithout having to manually edit files post-deployment.

**Why is this necessary:**

Since 'configure_misp.sh' runs **as part od the container entrypoint**, and the majority of users **pull the misp-core image directly** (instead of building it), it is currently **not possible to change this behavior easily** without:

- Overriding the entrypoint, or

- Mounting and overriding 'configure_misp.sh' via a volume in docker-compose.yml

Both approaches increase setup complexity unnecessarily. This PR aims to make the container **more configurable out-of-the-box** via environment variables, as intended in a Dockerized workflow.

**Benefits:**

- Avoids unintended JAR behavior caused by unnecessary client_secret
- Enables proper PKCE support via code_challenge_method
- Requires no volume overrides or post-deploy edits
- Futur-proof for public and private client use cases

**Not for users:**

In existing setups using prebuilt misp-core images, these changes my require:

- Rebuilding the image manually since 'configure_misp.sh' is baked into the image
- Or waiting for the next official image release to reflect the updated behavior

Let me know if you'd prefer this split into smaller commits or want updated documentation to go with it.

Thanks for maintaining a great project!


core/files/configure_misp.sh
```diff
@@ -95,6 +95,7 @@ set_up_oidc() {
                 ${OIDC_ISSUER:+\"issuer\": \"${OIDC_ISSUER}\",}
                 \"client_id\": \"${OIDC_CLIENT_ID}\",
                 \"client_secret\": \"${OIDC_CLIENT_SECRET}\",
+                \"code_challenge_method\": \"${OIDC_CODE_CHALLENGE_METHOD}\",
```
ostefano, line 98: should we now check this variable in the check above or empty values are the default values?

chaguser: Thanks for the question!
Regarding the code_challenge_method field:
When OIDC_CODE_CHALLENGE_METHOD is left empty (i.e., set but without a value), MISP defaults to using "plain" as the challenge method — which is also the current behavior without this PR (the default value).

However, the OIDC_CODE_CHALLENGE_METHOD variable just needs to be uncommented in the .env file, just like all other OIDC-related variables that are used in configure_misp.sh.

I just tested it earlier, if the OIDC_CHALLENGE_METHOD variable is empty, MISP falls back to its default behavior, exactly the same behavior before all these changes.

chaguser: If you wish, that the OIDC_CODE_CHALLENGE_METHOD takes the default value even in case it stayed as a comment on the .env file, let me know. I can add some changes

ostefano (approved):

PR #271, opened 2025-07-13
Update template.env

overcuriousity:
The value in OIDC_ROLES_MAPPING must not be escaped.

I had problems with OIDC failing and producing an access forbidden error when it was escaped, but the unescaped, unquoted integer value works.


ostefano:
LGTM

PR #273, opened 2025-07-15
Update Dockerfile

### LINKED ISSUE #265: /usr/local/bin/supervisord: No such file or directory
Hello,

I'm getting the below error when starting the containers.

```
/entrypoint.sh: line 80: /usr/local/bin/supervisord: No such file or directory
```

This is on a new RHEL88 machine, to avoid previous overlays/configs to get in the way. After installation Python 3.12, pip and podman were installed. It's using a default .env (copied from template.env, apart from changing the BASE_URL and MYSQL settings, and setting CORE_FLAVOR=full and MODULES_FLAVOR=full nothing is changed). Using the same .env file on Ubuntu24 builds just fine. Installation is done as root, to avoid permission issues.
It's using podman-compose version 1.4.0 (latest available is 1.4.1) and podman version 4.9.4-rhel.

From what I can tell from the build log, it's ignoring the "supervisor" (from https://github.com/MISP/misp-docker/blob/master/core/Dockerfile#L180). Seems other modules are added correctly, except not supervisor.

Building was done with `PODMAN_COMPOSE_VERBOSE=1 podman compose build | tee build.log`. After building the images, it also show some incomplete (?) images, but there were no -as far as I can check- build errors.

I'm a bit at a loss what's causing this problem. Do you have any good hints where to look for to solve this?

**Incomplete images**
```
REPOSITORY                             TAG                 IMAGE ID      CREATED       SIZE
ghcr.io/misp/misp-docker/misp-core     latest              27d2dcddf05d  10 hours ago  1.23 GB
<none>                                 <none>              0fef57aa5229  10 hours ago  1.74 GB
ghcr.io/misp/misp-docker/misp-modules  latest              aee43dcb3d9f  10 hours ago  1.42 GB
<none>                                 <none>              ca895e52b5d7  10 hours ago  876 MB
<none>                                 <none>              f33629ea38de  10 hours ago  624 MB
<none>                                 <none>              9e55a7ff8921  10 hours ago  381 MB
docker.io/valkey/valkey                7.2                 d679fcbee012  3 weeks ago   114 MB
```

**Commands issued afer installation**
```
subscription-manager repos --enable codeready-builder-for-rhel-8-$(arch)-rpms
dnf install git-core
dnf install python3.12
dnf install python3.12-pip.noarch
alternatives --config python3
pip3 install podman-compose
dnf install podman
```

**build.log** is attached
[build.log](https://github.com/user-attachments/files/21051047/build.log)

cudeso:
Fixes https://github.com/MISP/misp-docker/issues/265

Solves variable expansion differences between Podman and Docker.

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -165,7 +165,7 @@ FROM php-base AS python-build
         cd /var/www/MISP || exit; git submodule update --init --recursive .
 EOF
 
-    RUN <<-EOF
+    RUN bash <<'EOF'
         mkdir /wheels
 
         # Add additional dependencies (container specific)


ostefano:
I will need to investigate this more.

How do you run podman to build the images?
I would like to add more build tests

ostefano:
@cudeso have you tried using the SHELL variable for the whole build process (https://docs.docker.com/reference/dockerfile/#shell) ?

cudeso:
@ostefano 
Yes. 
When using `SHELL ["/usr/bin/sh", "-c"]` or `SHELL ["/usr/bin/bash", "-c"]` as the first line in `core/Dockerfile` the build completes fine, but the original error message with `/usr/local/bin/supervisord` re-occurs.

I've done each build with the same command, which basically resets the system to its original state.

`podman system prune ; podman image rm ghcr.io/misp/misp-docker/misp-core ; rm -f build.log ; cp template.env .env ; PODMAN_COMPOSE_VERBOSE=1 podman compose build --no-cache | tee build.log`

ostefano:
That is weird though right? We would expect setting bash globally for the build to fix this. Since it's a multi step build, have you tried adding it just to the offending step?

cudeso:
> That is weird though right? We would expect setting bash globally for the build to fix this. Since it's a multi step build, have you tried adding it just to the offending step?

Indeed. Don't fully understand why the global setting doesn't fix it.
I added it on the offending step with `RUN bash <<'EOF'` (the PR); I have not tried it with adding the SHELL command prior to that step, as from the documentation the SHELL command applies to everything that follows next in the Dockerfile, except if you override it with another shell.

core/Dockerfile
```diff
@@ -165,7 +165,7 @@ FROM php-base AS python-build
         cd /var/www/MISP || exit; git submodule update --init --recursive .
 EOF
 
-    RUN <<-EOF
+    RUN bash <<'EOF'
```
ostefano, line 168: @cudeso are quotes necessary here? My understanding is that they are not

ostefano (changes requested):
See comment.

core/Dockerfile
cudeso: I initially also considered them optional, but removing them results in build error

```
4/5] STEP 19/19: RUN bash <<-EOF
bash: line 18: [: : integer expression expected
Skipping overwriting  due to missing version variable
--> 518f13e9d985
--> 76e4f9f0a3dd
[5/5] STEP 11/23: RUN pip install --no-cache-dir /wheels/*.whl && rm -rf /wheels
WARNING: Requirement '/wheels/*.whl' looks like a filename, but the file does not exist

[notice] A new release of pip is available: 25.0.1 -> 25.1.1
[notice] To update, run: pip install --upgrade pip
ERROR: *.whl is not a valid wheel filename.
Error: building at STEP "RUN pip install --no-cache-dir /wheels/*.whl && rm -rf /wheels": while running runtime: exit status 1
Error: executing /usr/local/bin/podman-compose build --no-cache: exit status 1
```

ostefano: This is getting so random...

Do we know what was the shell expansion that did not work without this workaround?
Asking because we might need to retrofit this workaround to other build steps.

cudeso: Yes, basically the expansion did not work in the for loop (line 180 and further).
Printing out the variables `PYPI_MODULE_NAME_VERSION` before and after the the "for mod in" loop indicated they were set. But within the for loop they were empty.
In that for loop, it affected all `PYPI_` variables.

cudeso: It is indeed very weird/random. Using bash, with escaping the EOF was eventually the only way I found working to get those variables retain their value.

ostefano: Not happy yet, trying to reproduce things locally. How are you building it? Are you using podman-compose?

Because if I do so via `COMPOSE_PROGRESS=plain podman compose build misp-core`, I get this:

```
Skipping overwriting redis due to missing version variable
Skipping overwriting lief due to missing version variable
Skipping overwriting pydeep2 due to missing version variable
Skipping overwriting python-magic due to missing version variable
Skipping overwriting misp-lib-stix2 due to missing version variable
Skipping overwriting maec due to missing version variable
Skipping overwriting mixbox due to missing version variable
Skipping overwriting cybox due to missing version variable
Skipping overwriting pymisp due to missing version variable
Skipping overwriting misp-stix due to missing version variable
Adding missing module setuptools with version '==80.3.1'
Adding missing module supervisor with version '==4.2.5'
```

ostefano: Same with `podman-compose` (skipping the compose provider selection)

ostefano:
As per chat, this will be converted into a doc update where we suggest the workaround to users relying on an old (pre 5) version of `podman`

cudeso:
> As per chat, this will be converted into a doc update where we suggest the workaround to users relying on an old (pre 5) version of `podman`

Added https://github.com/MISP/misp-docker/pull/285

PR #279, opened 2025-07-17
Rebuilding MISP 2.5.15

HugeekBo:
Since MISP has release the new 2.5.16, we should build as well.


jeremiah-RENISAC:
Duplicate with PR #277

ostefano:
This has been handled in another PR

Issue #278, opened 2025-07-17
[BUG] cron sync job not working


LukeVader-IV:
I noticed that new events were not being syncronised to MISP.

Trying to debug, I entered the docker container with a bash shell.
Executing the command run by cron manually, i get the following error:
``` bash
root@487a39b0f6fd:/var/www/MISP# sudo -H -u www-data bash -c '/var/www/MISP/app/Console/cake Server pullAll 1'
Error: User ID do not match an existing user.
```

With a different UID, e.g. 0 for root:
``` bash
root@487a39b0f6fd:/var/www/MISP# /var/www/MISP/app/Console/cake Server pullAll 0
Usage: MISP/app/Console/cake Server pullAll [user_id] [full|update]
```
That message informing me about how it's used is odd, considering I did exactly what it's telling me to do?
As a sanity check, i tried with either the full or update args appended to the end, and the output was identical.

ostefano:
I believe the user id is the id of the admin user in MISP (usually 1 would be the correct, unless you deleted and added users). MISP developers would be able to confirm.

LukeVader-IV:
Oh all right i see. Yes i had done some experimenting with user accounts and auth, so the UID 1 no longer exists.
I didn't realize that the UID was tied to MISP users.
My applologies for my confusion. There is no documentation on this in the README besides a mention that it exists, something about it could be helpful.

LukeVader-IV:
The wording of `CRON_USER_ID` is unclear. maybe it should be called something along the lines of `SYNC_USER_ID` as the former indicates to me the UNIX UID that the cron task is run under.

ostefano:
You are right. If you have a spare cycle, can you update the doc and open a PR? Happy to merge it!

PR #280, opened 2025-07-21
replace CRON_USER_ID with SYNC_USER_ID

LukeVader-IV:
variable name CRON_USER_ID is confusing, replace it with SYNC_USER_ID
as discussed in #278

Warning: release will require a notice that the variable name has been changed


ostefano:
Can you implement a mechanism to grandfather existing setups?
Like reading the old value in case it exists...

ostefano:
@LukeVader-IV I will close this if the previous concern is not addressed.

Without a proper mechanism to retrofit existing installations, merging this would trigger a spike of Issues from users unaware of the change (bumping a tag in the docker-compose should not require reading a readme file).

PR #282, opened 2025-07-24
Support Selinux enabled platforms

trpyn:
Hello, I noticed, when using a Selinux enabled platform,  `docker compose up` fails to properly run the services with a `permission denied` error. 
This can be simply solved by adding the `:Z` tags to volumes mapping in `docker-compose.yml`.

I tested on Fedora42, it solved the issue for me.


ostefano:
@entropyqueen looks good.

I have two small requests:
- Can you add the `:Z` tag on the commented volumes as well?
- Can you add a paragraph in the documentation explaining the role of the `:Z` tag?

Thanks!

trpyn:
Here you go! 
However, the use of the `:Z` tag allows to give write permission to a volume bind, but I am unsure of which ones needs it and which don't. I'm willing to remove those that are unnecessary if you can point me which ones only requires read permissions

ostefano (approved):

ostefano (approved):

PR #290, opened 2025-08-14
Prevent writing plaintext admin key and password to stdout

jeremiah-RENISAC:
In kubernetes environments, everything written to stdout is captured and logged. This change prevents initialization from writing sensitive credentials out as cleartext


ostefano (changes requested):
Make this behavior optional (via ENV var and default to false) otherwise debugging new installations becomes even more challenging.

jeremiah-RENISAC:
> Make this behavior optional (via ENV var and default to false) otherwise debugging new installations becomes even more challenging.

Sure! Is it fine to make the default behaviour "Don't write in plaintext" with the envar to re-enable the cleartext credentials for a "secure by default" approach?

ostefano:
I would prefer not to.

ostefano (approved):

template.env
```diff
@@ -44,6 +44,9 @@ ADMIN_ORG_UUID=
 ADMIN_KEY=
 # defaults to MISP's default (admin)
 ADMIN_PASSWORD=
+# Prevent MISP Initialization from writing ADMIN_KEY and ADMIN_PASSWORD in plaintext (default: false)
+# Recommend setting to true in production or kubernetes environments where output is logged.
+DISABLE_PRINTING_PLAINTEXT_CREDENTIALS=
```
ostefano: Note that the default is not false, but empty. Also `false` would eval it the test to true.
Also, you need to update `docker-compose.yml` for the variable to be propagated.
I would comment this to true and keep it commented.

ostefano (changes requested):
See comment

ostefano:
@jeremiah-RENISAC CC

template.env
jeremiah-RENISAC: Awesome, thanks for catching that! I'll get this fixed (likely early next week) and resubmitted for review.

karollynecosta:
@jeremiah-RENISAC @ostefano This PR would help a lot. Let's proceed?

ostefano:
Waiting on @jeremiah-RENISAC to iterate

ostefano (approved):

PR #294, opened 2025-08-21
Fix misp-core shutdown (supervisord with exec as PID 1)

### LINKED ISSUE #293: misp-core exits with code 137 on shutdown due to missing exec in supervisord start
Currently, the container `misp-core` terminates on docker stop or on host reboot with `(exit code 137)`.
Other services such as `db` or `redis` shut down cleanly` (exit code 0).`

cause:
supervisord is not started via exec as PID 1 → Docker does not send SIGTERM correctly to the workers

The solution is already known and follows with a new pull request

HiS3:
## Problem

Currently, the container `misp-core` terminates on docker stop or on host reboot with `exit code 137.`
Other services such as `db` or `redis` shut down cleanly (exit code 0).

## Reason:
In the current core/Dockerfile supervisord is started without exec.
As a result, supervisord does not run as PID 1 in the container.
Signals `(SIGTERM,` `SIGINT)` that Docker sends to PID 1 are not forwarded correctly to the workers.

Result: _**Docker must exit hard after the stop timeout has expired → Exit 137.**_

## Solution
Start supervisord via exec so that it acts as PID 1.
Signals are thus correctly propagated to the workers managed by supervisor.
docker stop or docker-compose down then lead to a clean exit (0).

## Changes
core/Dockerfile: ENTRYPOINT adapted so that `exec /usr/bin/supervisord -c /etc/supervisor/supervisord.conf `is used.

## Tests
docker-compose build core
docker-compose up -d
docker stop -t 60 misp-core-1
→ Exit code 0, clean shutdown confirmed.

References:
Fixes #293


ostefano:
Love this, thank you.

Will merge after some testing.

ostefano:
Squashed and rebased and merged in another branch.

Issue #292, opened 2025-08-20
misp-docker relies on hard-coded cron jobs instead of new scheduled tasks system


AliceGrey:
The current docker configuration creates default cron jobs that should likely be replaced using the new scheduled tasks system introduced in v2.5.18

Cron jobs:
https://github.com/MISP/misp-docker/blob/678781632c0aab0fdbdcf599f40e252cf1664eac/core/files/entrypoint_cron.sh#L13-L26

2.5.18:
https://www.misp-project.org/2025/08/20/misp.2.5.18.released.html/

ostefano:
I believe a new background task has been added. @righel can you chime in here?
Should we remove those?

righel:
I'll start working on this soon

ostefano:
Done, new version (beta) is building right now

PR #296, opened 2025-08-21
Replace cron with tasks

righel:
replace cron jobs with the new scheduled tasks MISP feature.

solves: https://github.com/MISP/misp-docker/issues/292
requires: https://github.com/MISP/MISP/pull/10463


HiS3:
Wouldn't the [template.env](https://github.com/MISP/misp-docker/blob/replace-cron-with-tasks/template.env) have to be adapted accordingly?

```
# By default, a daily synchronization is performed, but you can modify this by changing the push and pull crontabs.
# CRON_PULLALL="0 1 * * *"
# CRON_PUSHALL="0 0 * * *"
```

righel:
> Wouldn't the [template.env](https://github.com/MISP/misp-docker/blob/replace-cron-with-tasks/template.env) have to be adapted accordingly?
> 
> ```
> # By default, a daily synchronization is performed, but you can modify this by changing the push and pull crontabs.
> # CRON_PULLALL="0 1 * * *"
> # CRON_PUSHALL="0 0 * * *"
> ```

good catch, thanks. updated it.

ostefano (approved):

ostefano:
Looks good but I would change INSERTs to INSERT ON DUPLICATE KEY UPDATE.

Also `updateGalaxies` appears twice.

PR #297, opened 2025-08-26
Refactor misp modules for nonroot

### LINKED ISSUE #291: misp-core and misp-modules non-root effective user
Similar and related to https://github.com/MISP/misp-docker/issues/133, but focused only on the user as root issue.
I'm have a requirement that the containers effective user must be non-root.
Is this doable?

eephillip:
A start to possibly allow misp-module as non-root. 

Some older prior art from way back ref:
https://github.com/MISP/misp-modules/blob/241824870e9a7057eaead167209c0e3d882e9ccf/docker/Dockerfile

Ref: https://github.com/MISP/misp-docker/issues/291

### DIFF
--- a/modules/Dockerfile
+++ b/modules/Dockerfile
@@ -1,4 +1,7 @@
 ARG DOCKER_HUB_PROXY=""
+ARG MISP_UID=1001
+ARG MISP_GID=1001
+
 
 FROM "${DOCKER_HUB_PROXY}python:3.12-slim-bookworm" AS python-build
     ENV DEBIAN_FRONTEND=noninteractive
@@ -16,7 +19,6 @@ FROM "${DOCKER_HUB_PROXY}python:3.12-slim-bookworm" AS python-build
             apt-get install -y --no-install-recommends \
                 ca-certificates \
                 git
-                
         else
             apt-get install -y --no-install-recommends \
                 ca-certificates \
@@ -59,6 +61,8 @@ EOF
 FROM "${DOCKER_HUB_PROXY}python:3.12-slim-bookworm"
     ENV DEBIAN_FRONTEND=noninteractive
     ARG MODULES_FLAVOR
+    ARG MISP_UID
+    ARG MISP_GID
 
     RUN <<-EOF
         apt-get update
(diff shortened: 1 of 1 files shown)


modules/Dockerfile
```diff
@@ -59,6 +61,8 @@ EOF
 FROM "${DOCKER_HUB_PROXY}python:3.12-slim-bookworm"
     ENV DEBIAN_FRONTEND=noninteractive
     ARG MODULES_FLAVOR
+    ARG MISP_UID
```
ostefano, line 64: You need to init them again

modules/Dockerfile
```diff
@@ -74,8 +78,13 @@ FROM "${DOCKER_HUB_PROXY}python:3.12-slim-bookworm"
         apt-get autoremove -y && apt-get clean -y && rm -rf /var/lib/apt/lists/*
 EOF
 
+    RUN groupadd -g "${MISP_GID}" misp; \
+        useradd  -u "${MISP_UID}" -g "${MISP_GID}" -m -d /home/misp -s /usr/sbin/nologin misp
```
ostefano, line 82: split in two RUN here

modules/Dockerfile
```diff
@@ -74,8 +78,13 @@ FROM "${DOCKER_HUB_PROXY}python:3.12-slim-bookworm"
...
+
     COPY --from=python-build /wheels /wheels
     RUN pip install --no-cache-dir /wheels/*.whl && rm -rf /wheels
-    RUN bash -c 'mkdir -p /custom/{action_mod,expansion,export_mod,import_mod}'
+    RUN bash -c 'mkdir -p /custom/{action_mod,expansion,export_mod,import_mod} && chown -R misp:misp /custom'
```
ostefano, line 86: Why && and not a new RUN command?

ostefano (changes requested):

modules/Dockerfile
eephillip: ah, yeah. Intended to move these up to the global scope, will adjust

modules/Dockerfile
eephillip: I'm combining both of these statements solely for the purpose of a adhering the best practices of minimizing diff layers from each RUN directive

modules/Dockerfile
eephillip: I'm combining both of these statements solely for the purpose of a adhering the best practices of minimizing diff layers from each RUN directive

PR #300, opened 2025-09-08
Allow passwordless Redis connections

### LINKED ISSUE #299: Redis configuration doesn't support password-less redis
I'm using the image in a Kubernestes environement where the Redis services is protected at the network level and doesn't required auhtentication from the apps in the same namespace. 

Curently the entrypoint script enforces the "redispassword" when the REDIS_PASSWORD environment variables isn't present or is set to the empty string. https://github.com/MISP/misp-docker/blob/master/core/files/entrypoint.sh#L18

Here's an example if my environment configuration regading  Redis:
```
root@misp-core-64ff664858-57f6z:/var/www/MISP/app/Config# printenv | grep REDIS
REDIS_PASSWORD=
REDIS_HOST=misp-redis-standalone.misp.svc.cluster.local
REDIS_PORT=6379
```

Here the produced config.php
```
root@misp-core-64ff664858-57f6z:/var/www/MISP/app/Config# cat config.php | grep redis
    'redis_host' => 'misp-redis-standalone.misp.svc.cluster.local',
    'redis_password' => 'redispassword',
    'redis_port' => 6379,
    'redis_database' => 13,
    'ZeroMQ_redis_host' => 'misp-redis-standalone.misp.svc.cluster.local',
    'ZeroMQ_redis_password' => 'redispassword',
    'ZeroMQ_redis_port' => 6379,
    'redis_host' => 'misp-redis-standalone.misp.svc.cluster.local',
    'redis_password' => 'redispassword',
    'redis_port' => 6379,
    'redis_database' => 1,
    'redis_namespace' => 'background_jobs',
```

I think for the docker-compose deployment where the Redis service is configured to have a password by default, it makes sense to have that default value. But the entrypoint shouln't enforce a default password or at least should allow passwordless configuration for other kind of deployments.

damanzanofortra:
Potential solution for https://github.com/MISP/misp-docker/issues/299


ostefano (changes requested):
I would allow a guard, like a variable REDIS_ENABLE_EMPTY_PASSWORD that if true will allow empty password. Maybe ENABLE_REDIS_EMPTY_PASSWORD better.

damanzanofortra:
Perfect, let me implement that suggestion

damanzanofortra:
@ostefano just added the requested guard, and update local dev tools in case someone wants to use the local redis passwordless. Please let me know if there's anything else that should be updated.

ostefano (commented):
Looks good, but will need to test it a bit

ostefano (approved):

PR #301, opened 2025-09-12
Add MISP-Guard container

### LINKED ISSUE #22: Feature request: Add misp-guard docker to potifolio
Having this repo also build a image for misp-guard would be nice - https://github.com/MISP/misp-guard/

monjiapawne:
This PR adds MISP-Guard as an optional service.  
It's disabled by default but can be toggled with:
```
COMPOSE_PROFILES=misp-guard
```
Notes:
- Added the `entrypoint.sh` to the service so there was a way to dynamically target the `misp-core` container IP from docker's DNS.
- Opted for `depends_on` at misp-core start since I've had misp-core flap on first run.  
- `healthcheck` for misp-guard is set to `2m`, to avoid spamming the logs.
- The `README` has all the configuration is greater detail.

I've tested this extensively with clean builds, functions exactly as an external misp-guard would.  

Current dockerfile pulls misp-guard via git clone at build time, ideally this should be replaced with a published image and pinned version.

Fixes #22


ostefano:
Love this!

May I suggest to also extend the GitHub actions so we build the image?

monjiapawne:
Thanks @ostefano, I've extended the test and release workflows to include misp-guard

ostefano:
Cool. Will merge as soon as I have a spare cycle to test it.

Thanks again!

ostefano (changes requested):
Like we do with misp-core and misp-modules the commit should just be a fallback in case the version tag is not available.

Also, template.env should have the version tag initialized.

After these two concerns are addressed, I will merge 👍

ostefano:
I am doing a pass because as things they can not work (for example, they require cloning the repo for using it).

I will ask you to reivew it

ostefano (approved):

ostefano (approved):

PR #305, opened 2025-09-17
Add persistence to Valkey

### LINKED ISSUE #298: Enabled modules are not persistent
We've deployed MISP in our project and have noticed that the enabled status of our modules is not being saved. After enabling a module through the UI and then restarting the Redis service, the module's status reverts to disabled. This is causing issues with our automated workflows. We're looking for a way to persist these settings.

Steps to Reproduce
Deploy MISP

Manually restart the Redis service.

Navigate back to the worker settings.

Expected Behavior
The module should remain enabled after the Redis restart.

Actual Behavior
The module's status reverts to disabled.

We've observed that the enabled status of modules seems to be stored in the cache, unlike the configuration for triggers.(The trigger settings are persistent) We lose this setting every time  when we restarted for maintenance. Is there a way to make these settings persistent?

claudex:
Valkey is [configured by default][1] to persist data on disk, so we only need the volume configure to keep it between container recreation. The container has already [the required directory][2].

Fixes #298

[1]: https://valkey.io/topics/persistence/#no-persistence
[2]: https://hub.docker.com/r/valkey/valkey/#start-with-persistent-storage

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,6 +20,8 @@ services:
       retries: 3
       start_period: 5s
       start_interval: 5s
+    volumes:
+      - cache_data:/data:Z
 
   db:
     # We use MariaDB because it supports ARM and has the expected collations
@@ -277,4 +279,5 @@ services:
 
 volumes:
     mysql_data:
+    cache_data:
 


ostefano:
~~I remember we disabled it for a reason, but I can't recall the reason at the moment 😅~~

I see the use case here: https://github.com/MISP/misp-docker/issues/298

ostefano (approved):

PR #307, opened 2025-09-18
Improve errror message for env file

claudex:
I got confused when it said that the .env file was missing but it was just missing the new var from misp-guard. I think this message will avoid some confusion for users.


claudex:
Sorry mixed 2 changes in one commit by mistake, will split them and resubmit each separately

PR #310, opened 2025-09-18
Allow to configure the exposed port for the core container

claudex:
Simple config to allow other exposed port than default one (like for a rootless podman use).

Next steps could be to have other port inside the container and automatically update the base url if possible. But I need to check the impact.

### DIFF
--- a/template.env
+++ b/template.env
@@ -58,6 +58,9 @@ CRON_USER_ID=
 # defaults to 'https://localhost'
 # note: if you are exposing MISP on a non-standard port (i.e., the port is part of the URL you would use to access it, e.g., https://192.168.0.1:4433) you need to include the port in the BASE_URL variable
 BASE_URL=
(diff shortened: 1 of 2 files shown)


template.env
```diff
@@ -58,6 +58,9 @@ CRON_USER_ID=
 # defaults to 'https://localhost'
 # note: if you are exposing MISP on a non-standard port (i.e., the port is part of the URL you would use to access it, e.g., https://192.168.0.1:4433) you need to include the port in the BASE_URL variable
 BASE_URL=
+# defaults to 80 and 443, don't forget to update the base url if not the defaults one
+CORE_HTTP_PORT=
```
ostefano: Comment these ones

ostefano (changes requested):

ostefano (approved):

PR #269, opened 2025-07-10
Allow splitting image to multiple containers for Kubernetes deployment

jeremiah-RENISAC:

PR #311, opened 2025-09-27
Kubernetes specific entrypoints with deployment/kustomize examples

oivindoh:
Based on #203 and #269

Implements two new entrypoints for use in kubernetes land in order to be able to separate misp and nginx.

None of the changes should affect/break existing instances in or out of kubernetes.

- [Deployment examples](https://github.com/oivindoh/misp-docker/blob/k8s/kubernetes/README.md) for misp, redis and mysql are provided
- [Replacement CronJobs](https://github.com/oivindoh/misp-docker/blob/k8s/kubernetes/cronjobs/) making old cron jobs and new Scheduled Tasks unnecessary. Insertion of these scheduled tasks is thus disabled if kubernetes mode is active.
- A [cilium network policy example](https://github.com/oivindoh/misp-docker/blob/k8s/kubernetes/manifests/policy-cilium.yaml) is also provided.

Build manifests using `kustomize build kubernetes/`

Additional changes related to kubernetes (because pod logs tend to be shipped to systems accessed by people who should not be privy to system secrets) but not exclusive to kubernetes is:

- Expansion of the existing DISABLE_PRINTING_PLAINTEXT_CREDENTIALS environment variable to suppress printing certain sensitive settings set by utility functions, as well as redis password.
- Supervisord worker logs to container log
- error.log to container logs


oivindoh:
Realising I haven't handled files here, as we use these includes:

```yaml
  s3.defaults.json: |
    {
        "Plugin.S3_enable": {
          "default_value": true
        },
        "Plugin.S3_aws_compatible": {
          "default_value": true
        },
        "MISP.attachments_dir": {
          "default_value": "s3://"
        }
    }
  s3.envars.json: |
    {
      "Plugin.S3_bucket_name": {
        "default_value": "${S3_BUCKET}"
      },
      "Plugin.S3_aws_endpoint": {
        "default_value": "${S3_ENDPOINT}"
      },
      "Plugin.S3_aws_access_key": {
        "default_value": "${S3_ACCESS_KEY}"
      },
      "Plugin.S3_aws_secret_key": {
        "default_value": "${S3_SECRET_KEY}"
      },
      "MISP.attachments_dir": {
        "default_value": "s3://"
      }

```

core/files/entrypoint.sh
```diff
@@ -84,5 +84,22 @@ export NGINX_X_FORWARDED_FOR=${NGINX_X_FORWARDED_FOR:-false}
...
+            echo " - NGINX"
+            exec /kubernetes/entrypoint_nginx.sh
+        ;;
+        php*)
+            echo " - PHP (MISP)"            # Not ideal, but let supervisord manage the workers still
+            mv /etc/supervisor/conf.d/10-supervisor.conf{.kubernetes,}
+            /usr/local/bin/supervisord -c /etc/supervisor/supervisord.conf &
+            exec /kubernetes/entrypoint_fpm.sh
+        ;;
+    esac
```
claudex: I don't understand why this case switch here is useful. Shouldn't be easier to set directly the entrypoint in the kubernetes container definition (in the deployment) since it should require a startup args in any case ?

core/files/entrypoint_fpm.sh
```diff
@@ -8,6 +8,27 @@ term_proc() {
...
+    done)
+
+    if [[ -n "$stdout_fd" ]]; then
+        :
+    else
+        stdout_fd=1
+    fi
+
+    tail -F /var/www/MISP/app/tmp/logs/error.log > /proc/1/fd/${stdout_fd} 2>/dev/null &
+}
```
claudex, line 13: I don't understand why this complicated search for stdout. It seams fragile and since it's not rewritten in the app, the standard stdout fd should work with better robustness.

kubernetes/mysql-credentials.env
```diff
@@ -0,0 +1,2 @@
+MYSQL_PASSWORD=sometimespasswordsarelongsometimespasswordsareshort
+MYSQL_ROOT_PASSWORD=sometimespasswordsarelongsometimespasswordsareshort
```
claudex, line 2: I think it should be named mysql-credentials-template.env

core/files/etc/supervisor/conf.d/50-workers.conf
```diff
@@ -10,9 +10,11 @@ process_name=%(program_name)s_%(process_num)02d
...
 autostart=false
 autorestart=true
-redirect_stderr=false
-stderr_logfile=/var/www/MISP/app/tmp/logs/misp-workers-errors.log
-stdout_logfile=/var/www/MISP/app/tmp/logs/misp-workers.log
+redirect_stderr=true  ; Changed: Merge stderr into stdout for simpler aggregation (optional; see notes below)
+stdout_logfile=/dev/stdout  ; Changed: From file path
+stdout_logfile_maxbytes=0  ; Changed: Disable rotation
+stderr_logfile=/dev/stderr  ; Changed: From file path (only needed if redirect_stderr=false)
+stderr_logfile_maxbytes=0  ; Changed: Disable rotation
```
claudex: I'm not sure we want to have this change for all deployments by default

core/files/entrypoint_fpm.sh
```diff
@@ -30,14 +51,19 @@ change_php_vars() {
...
+                echo "Configure PHP | Setting 'session.save_path = '$(echo $REDIS_HOST | grep -E '^\w+://' || echo tcp://$REDIS_HOST):$REDIS_PORT?auth=${ESCAPED}'"
+            fi
             sed -i "s|.*session.save_path = .*|session.save_path = '$(echo $REDIS_HOST | grep -E '^\w+://' || echo tcp://$REDIS_HOST):$REDIS_PORT?auth=${ESCAPED}'|" "$FILE"
         else
             echo "ERROR: REDIS_PASSWORD is not set but ENABLE_REDIS_EMPTY_PASSWORD is false. Please set REDIS_PASSWORD or enable ENABLE_REDIS_EMPTY_PASSWORD=true for passwordless Redis."
             exit 1
         fi
         sed -i "s/session.sid_length = .*/session.sid_length = 64/" "$FILE"
         sed -i "s/session.use_strict_mode = .*/session.use_strict_mode = 1/" "$FILE"
+        sed -i "s|session.cookie_domain = .*|session.cookie_domain = ${BASE_URL}|" "$FILE"
```
claudex: I don't think we should have this by default for all instances since it could break existing ones.

core/files/etc/supervisor/conf.d/50-workers.conf
oivindoh: Could split it out in its own PR I guess, but it makes little sense to keep the current paths in a container environment

core/files/entrypoint_fpm.sh
oivindoh: I agree

kubernetes/mysql-credentials.env
oivindoh: I don't agree - I want this to be consumable as-is as a base resource for kustomize overlays, and having the copy template -> edit copy workflow used by the docker-compose setup would break that.

core/files/entrypoint.sh
oivindoh: We could always set the entrypoint instead of args and source this current entrypoint for the vars, I just don't really see the advantage vs reacting to arg?

core/files/etc/supervisor/conf.d/50-workers.conf
claudex: I agree that's not the best practice for containers, but changing it will potentially break existing deployment. So I think it should be parametrized depending on the expected setup.

core/files/entrypoint_fpm.sh
oivindoh: This insanity is left over from a power struggle between me and reality. Thanks for catching it 👍

core/files/etc/supervisor/conf.d/50-workers.conf
oivindoh: I'll make it k8s only

core/files/etc/supervisor/conf.d/10-supervisor.conf.kubernetes
adulau, line 1: Can you remove those default username/password?

core/files/entrypoint.sh
oivindoh: On second thought this is indeed cleaner. Pushed necessary changes now.

core/files/etc/supervisor/conf.d/50-workers.conf
oivindoh: Now supplying a k8s specific worker config like for the main config

core/files/etc/supervisor/conf.d/10-supervisor.conf.kubernetes
oivindoh: It's mainly to avoid an error message as things stand, and it's the same as for the non-kubernetes supervisor conf, but I agree it should be env var based instead of hardcoded

oivindoh: made it configurable, defaulting to the same as before to avoid breaking anything

core/files/configure_misp.sh
```diff
@@ -603,7 +611,11 @@ echo "MISP | Set Up Session ..." && set_up_session
 
 echo "MISP | Set Up Proxy ..." && set_up_proxy
 
-echo "MISP | Create default Scheduled Tasks ..." && create_default_scheduled_tasks
+if [[ -n "$KUBERNETES_SERVICE_HOST" && $# -gt 0 ]]; then
+    echo "MISP | Skipping Scheduled Tasks creation in kubernetes"
```
ostefano: Why would we want to do this?

core/files/entrypoint.sh
```diff
@@ -84,5 +84,13 @@ export NGINX_X_FORWARDED_FOR=${NGINX_X_FORWARDED_FOR:-false}
 export NGINX_SET_REAL_IP_FROM=${NGINX_SET_REAL_IP_FROM}
 export NGINX_CLIENT_MAX_BODY_SIZE=${NGINX_CLIENT_MAX_BODY_SIZE:-50M}
 
+export SUPERVISOR_HOST=${SUPERVISOR_HOST:-127.0.0.1}
+export SUPERVISOR_USERNAME=${SUPERVISOR_USERNAME:-supervisor}
+export SUPERVISOR_PASSWORD=${SUPERVISOR_PASSWORD:-supervisor}
+
+if [ -n "${BASH_SOURCE[0]}" ]; then
```
ostefano, line 92: Please add a comment here

core/files/entrypoint_fpm.sh
```diff
@@ -62,16 +70,26 @@ change_php_vars() {
...
         fi
+        if [[ -n "$PHP_LISTEN_FPM" ]]; then
+            echo "Configure PHP | Setting 'listen' to [::]:9002"
+            sed -i "/^listen =/s@=.*@= [::]:9002@" "$FILE"
+        fi
+
     done
 }
 
+if [ -n "${BASH_SOURCE[0]}" ]; then
```
ostefano, line 89: Same, comment needed

core/files/entrypoint_nginx.sh
```diff
@@ -255,7 +255,7 @@ flip_nginx() {
     echo "... nginx docroot set to ${NGINX_DOC_ROOT}"
     sed -i "s|root.*var/www.*|root ${NGINX_DOC_ROOT};|" /etc/nginx/includes/misp
 
-    if [[ "$reload" = "true" ]]; then
+    if [[ "$reload" = "true" ]] && [[ -z "$KUBERNETES_SERVICE_HOST" ]]; then
         echo "... nginx reloaded"
         nginx -s reload
```
ostefano, line 260: use `flip_nginx` (with the right arguments)

core/files/utilities.sh
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+sensitive_settings=(
+    "MISP.redis_password"
+    "Plugin.ZeroMQ_redis_password"
+    "SimpleBackgroundJobs.redis_password"
+    "Plugin.S3_aws_secret_key"
+    "Security.encryption_key"
+    "GnuPG.password"
+    )
```
ostefano: Indent

template.env
```diff
@@ -97,6 +97,11 @@ SYNCSERVERS_1_KEY=
 # 	SYNCSERVERS_1_PULL_RULES='{\"tags\":{\"OR\":[],\"NOT\":[]},\"orgs\":{\"OR\":[],\"NOT\":[]},\"url_params\":\"{\\\"searchanalysis\\\": \\\"2\\\"}\"}'
 SYNCSERVERS_1_PULL_RULES=
 
+# Supervisord host for external jobs
```
ostefano: external jobs?

ostefano (changes requested):
Just some minor comments.

I love this PR, let's keep up the momentum until we merge it!

template.env
oivindoh: prepares the ground for running jobs in another container

core/files/configure_misp.sh
oivindoh: I can remove this.  The rationale is because we run the supplied CronJobs in kubernetes instead

core/files/entrypoint.sh
oivindoh: comment added

core/files/entrypoint_fpm.sh
oivindoh: comment added

core/files/configure_misp.sh
ostefano: But they are not cron jobs anymore, they are handled by MISP background jobs system (scheduler worker)

core/files/utilities.sh
oivindoh: Removed indent  on line 10

core/files/entrypoint_nginx.sh
oivindoh: Reverted this entire change, it's unnecessary

template.env
oivindoh: edited the template comment a bit to clarify

core/files/kubernetes/entrypoint_fpm.sh
```diff
@@ -0,0 +1,37 @@
...
+if [[ -x /custom/files/customize_misp.sh ]]; then
+    echo "INIT | Customize MISP installation ..."
+    /custom/files/customize_misp.sh
+fi
+
+echo "Configure PHP | Change PHP values ..." && change_php_vars
+
+echo "Configure PHP | Starting PHP FPM"
+
+exec /usr/sbin/php-fpm8.2 -R -F
```
claudex: I think it would be better to use tini as PID 1, it would have better signal handling, process termination and enter the container.

core/files/kubernetes/entrypoint_nginx.sh
```diff
@@ -0,0 +1,15 @@
...
+# Initialize nginx
+echo "INIT | Initialize NGINX ..." && init_nginx
+
+echo "... setting 'fastcgi_pass' to misp-php:9002"
+sed -i "s@fastcgi_pass .*;@fastcgi_pass misp-php:9002;@" /etc/nginx/includes/misp
+
+echo "INIT | Flip NGINX live ..." && flip_nginx true false
+
+# launch nginx as current shell process in container
+exec nginx -g 'daemon off;'
```
claudex: Here also we sould use tini.

kubernetes/dhparams.pem
```diff
@@ -0,0 +1,13 @@
...
+nCCxjywx3QfB5vqXryqbe0DN4hT9G5J/15z3sh5NZdLNdn/tapkh+MJWRKzg06Tf
+mbe3xdcOIjJDYqa5XGYXzry4JabhNjRdBsLc1opqQGRAWijga49+VLB7YTnZgWS8
+64CN9l35r0Ui/v5bxM6QUi9Rp1QM+jzBhow1uRGhrpocVgLBTD+gTKX8f3YX5P7G
+Yplxw8KRgK3hXuyiKx2Fht+GcB4X68y8jxUgCufUi+L4SQm3CMwmg+KFSilOnUfZ
+XLEHdD0ZAUxR/p30LCEManKwpZSxfJbUd+O2MjkYa5/OZOlwESjQ4LOtSTH5Kvpd
+lcqCMA36p+GlOo/j/CzGCBLFSTYNlEEl56d6JU63xMyK8a7z9JkQOR8tPgCLe1Bw
+I7dv9TYMgIBmRS70Z442A1k2T5rrBJHG3UD5QouHoApqPme650yjWdCW4V4ma1xb
+b7Tfq93xtNpTTccxIMAdDSIncnnmGBd8p8LOc8fjBB/N29hn14GMmIcVcU0k961v
+Z0Q1d2ED8zsyJzYf3PL84sHZyUcEohjsgktESFx83Sxqs53CQLJ7xp8CAQICAgFF
+-----END DH PARAMETERS-----
```
claudex: I don't think we should have the dh parameters in the git repository. I know it is easier to deploy, and this is only use in internal traffic with the provided configuration but it is bad practice to use a public one and this could easily find its way in production.

docker-compose.yml
```diff
@@ -215,6 +215,11 @@ services:
       - "PROXY_METHOD=${PROXY_METHOD}"
       - "PROXY_USER=${PROXY_USER}"
       - "PROXY_PASSWORD=${PROXY_PASSWORD}"
+      # s3 settings
+      - "S3_BUCKET=${S3_BUCKET}"
+      - "S3_ENDPOINT=${S3_ENDPOINT}"
+      - "S3_ACCESS_KEY=${S3_ACCESS_KEY}"
+      - "S3_SECRET_KEY=${S3_SECRET_KEY}"
```
ostefano, line 222: Can you also add the SUPERVISOR env vars, otherwise they would not get propagated

ostefano (changes requested):

docker-compose.yml
oivindoh: good catch - added

core/files/entrypoint_fpm.sh
```diff
@@ -62,16 +70,27 @@ change_php_vars() {
             echo "Configure PHP | Setting 'pm.status_path = /status'"
             sed -i -E "s/;?pm.status_path = .*/pm.status_path = \/status/" "$FILE"
             echo "Configure PHP | Setting 'pm.status_path = /run/php/php-fpm-status.sock'"
+            # TODO: Respect version number in path?
```
ostefano: What do you mean here?

ostefano (approved):

core/files/entrypoint_fpm.sh
oivindoh: I was thining of making the socket path reflect the version path indicated by the file we're iterating through, but it's an unnecessary change anyway - deleted the comment

adulau:
This contribution is great, and we have many users who would be interested in the topic. I would love to see a blog post about it on the MISP Project website. If you feel comfortable with the idea, feel free to make a pull request on the MISP Project website repository: https://github.com/MISP/misp-website

core/files/etc/supervisor/supervisord.conf
```diff
@@ -3,11 +3,17 @@
 [unix_http_server]
 file=/var/run/supervisor.sock   ; (the path to the socket file)
 chmod=0700                       ; sockef file mode (default 0700)
+nodaemon=true
+username=%(ENV_SUPERVISOR_USERNAME)s
+password=%(ENV_SUPERVISOR_PASSWORD)s
 
 [supervisord]
-logfile=/var/log/supervisor/supervisord.log ; (main log file;default $CWD/supervisord.log)
+logfile=/dev/null ; (main log file;default $CWD/supervisord.log)
```
ostefano: Do we have an alternative here so we don't lose these logs?

Or conversely, why are these not important?

core/files/etc/supervisor/supervisord.conf
```diff
@@ -3,11 +3,17 @@
...
 pidfile=/var/run/supervisord.pid ; (supervisord pidfile;default supervisord.pid)
 childlogdir=/var/log/supervisor            ; ('AUTO' child log dir, default $TEMP)
+loglevel=error
```
claudex: I don't think this should be enforced by default

kubernetes/dhparams.pem
ostefano: +1

oivindoh: I don't quite understand the rationale for not including it? This is not sensitive data.

core/files/kubernetes/entrypoint_nginx.sh
oivindoh: moved to tini

core/files/kubernetes/entrypoint_fpm.sh
oivindoh: moved to tini

core/files/etc/supervisor/supervisord.conf
oivindoh: It's a change that's left behind from when I was tailing debug.log to stdout to avoid duplicates. It shouldn't be there, thanks for catching

core/files/etc/supervisor/supervisord.conf
oivindoh: Yes, this was an error; https://github.com/MISP/misp-docker/pull/311#discussion_r2398999866

kubernetes/dhparams.pem
ostefano: You are right, but could we keep the logic to generate them if not present?

oivindoh: Changed to the recommended standard group 16 parameters now. I think perhaps this should be built into the container itself though, rather than be part of the kubernetes specific setup.

oivindoh: (re: sensitivity: https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange#Secrecy_chart - the parameters in this PEM file are p + g)

claudex: It's not private. But if you can attack a specific params file to reach multiple target, this increase the price you can set for the attack. That's the issue that initiated logjam attack https://en.wikipedia.org/wiki/Logjam_(computer_security). So we want to avoid having multiple users of the same params

oivindoh: I guess you could argue that, but you could also argue 4096 is somewhat of a different beast to attack (thus openssh is happily supplying 2048 parameters, and windows uses 2048/4096 groups by default), and if someone with those kinds of resources / hitherto unknown tech want you, they can just grab your parameters and go at it anyway.

I think the better option is removing the DHE ciphers from nginx config entirely, so I made that change now.

claudex (approved):

core/files/etc/supervisor/conf.d/10-supervisor.conf.kubernetes
adulau: Thanks!

adulau (approved):
Thanks for the updates and fixes. Well done.

core/files/etc/misp-docker/s3.envars.json
```diff
@@ -0,0 +1,18 @@
...
+    "Plugin.S3_aws_endpoint": {
+        "default_value": "${S3_ENDPOINT}"
+    },
+    "Plugin.S3_aws_access_key": {
+        "default_value": "${S3_ACCESS_KEY}"
+    },
+    "Plugin.S3_aws_secret_key": {
+        "default_value": "${S3_SECRET_KEY}"
+    },
+    "MISP.attachments_dir": {
```
oivindoh, line 14: This feels a bit hacky to have here. Does anyone have a better idea?

ostefano: Do you mean the location of this setting?

PR #324, opened 2025-11-04
Ensure core MISP services always restart

mortezamirkar:
- add `restart: always` to the mail, redis, misp-core, misp-modules, and misp-guard services
- keep the stack resilient by restarting these containers after failures or daemon restarts


ostefano:
Does 'restart' result into starting the containers that are not explicitly flagged as dependency?

The scenario being: `docker compose up misp-core`.

mortezamirkar:
No @ostefano .
docker compose restart, only restarts the specified container(s) — it doesn’t start other containers or dependencies that aren’t already running.

ostefano (approved):

PR #326, opened 2025-11-04
Improve entrypoint sourcing guard and admin key defaults

mortezamirkar:
export AUTOGEN_ADMIN_KEY with a sensible default that matches AUTOCONF_ADMIN_KEY
tighten the entrypoint scripts’ sourcing check so they only return when actually sourced
make rest_client.sh exit with a non-zero status when jq or curl is unavailable


ostefano:
@oivindoh summoning you here, since the diff affects your bash magic.

core/files/entrypoint.sh
```diff
@@ -90,7 +91,7 @@ export SUPERVISOR_USERNAME=${SUPERVISOR_USERNAME:-supervisor}
 export SUPERVISOR_PASSWORD=${SUPERVISOR_PASSWORD:-supervisor}
 
 # Hinders further execution when sourced from other scripts
-if [ -n "${BASH_SOURCE[0]}" ]; then
```
ostefano, line 93: @oivindoh please review

core/files/entrypoint_fpm.sh
```diff
@@ -86,7 +86,7 @@ change_php_vars() {
 }
 
 # Hinders further execution when sourced from other scripts
-if [ -n "${BASH_SOURCE[0]}" ]; then
```
ostefano, line 89: @oivindoh please review

core/files/entrypoint_nginx.sh
```diff
@@ -395,7 +395,7 @@ init_nginx() {
 }
 
 # Hinders further execution when sourced from other scripts
-if [ -n "${BASH_SOURCE[0]}" ]; then
```
ostefano, line 398: @oivindoh please review

ostefano (commented):
Need @oivindoh

mortezamirkar:
@oivindoh

ostefano (approved):

PR #332, opened 2025-11-20
Update container image version in Kubernetes manifests and remove non-existent configmap reference

dakota-marshall:
Hey There!

We are working on deploying MISP into our Kubernetes cluster, and using the manifests in this repo as an example, we ran into two issues with them that I figure id contribute upstream:

- First, the manifests set a custom entry point for the php and nginx container of `/kubernetes/entrypoint_fpm.sh` and `/kubernetes/entrypoint_nginx.sh` respectively, but the image version used in the example does not have these files inside the container. I updated to `v2.5.25` and that resolved the problem
- The MISP NGINX manifest also tries to mount a `dhparams` configMap that does not seem to exist. There is no mention of this needing to be generated in the docs, so I commented it out.

Let me know if I need to make any adjustments, thanks!

### DIFF
--- a/kubernetes/manifests/deployment-misp.yaml
+++ b/kubernetes/manifests/deployment-misp.yaml
@@ -24,7 +24,7 @@ spec:
         - name: misp
           command: 
             - /kubernetes/entrypoint_fpm.sh
-          image: "ghcr.io/misp/misp-docker/misp-core:v2.5.25"
+          image: "ghcr.io/misp/misp-docker/misp-core:latest"
           imagePullPolicy: IfNotPresent
           readinessProbe:
             tcpSocket:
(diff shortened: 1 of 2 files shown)


ostefano:
LGTM

You can delete the commented lines (we don't use dhparam anymore).

Also, would it be OK to replace the version with the latest tag?

Otherwise it would be nice if it could read it from .env (even better).

dakota-marshall:
I went ahead and changed the image tags to latest and removed the commented config.

I am not super familiar with how to template out the image tag with Kustomize, I can do it easily in Helm, but not Kustomize, unfortunately. So I just set the tag to latest

ostefano:
LGTM

PR #333, opened 2025-11-23
Upgrade to debian trixie

ostefano:
Replace bookworm with trixie


ostefano:
@iglocska @oivindoh @UFOSmuggler @claudex @mortezamirkar please test this new version (start from a blank `.env` file)

ostefano:
Upgraded `misp-guard` as well. @monjiapawne please test this.

core/files/entrypoint.sh
```diff
@@ -12,7 +12,7 @@ export MYSQL_PORT=${MYSQL_PORT:-3306}
 export MYSQL_USER=${MYSQL_USER:-misp}
 export MYSQL_PASSWORD=${MYSQL_PASSWORD:-example}
 export MYSQL_DATABASE=${MYSQL_DATABASE:-misp}
-export MYSQL_CMD="mysql -u $MYSQL_USER -p$MYSQL_PASSWORD -P $MYSQL_PORT -h $MYSQL_HOST -r -N $MYSQL_DATABASE"
```
ostefano, line 15: @UFOSmuggler I would like this switch to be used only when MYSQL_TLS is true. Can you add a commit with this? Thanks!

ostefano: Maybe:

```
${MYSQL_TLS:+}${MYSQL_TLS:- --skip-ssl}
```

?

UFOSmuggler: can i ask why?  i may misunderstand what you're trying to do.  the below misp instance uses and requires TLS which won't work with `--skip-ssl`:

```
root@ip-192-168-2-216:/var/www/MISP# mysql -u $MYSQL_USER -p$MYSQL_PASSWORD -P $MYSQL_PORT -h $MYSQL_HOST -r -N $MYSQL_DATABASE --skip-ssl
ERROR 3159 (08004): Connections using insecure transport are prohibited while --require_secure_transport=ON.
```

same host without `--skip-ssl`

```
root@ip-192-168-2-216:/var/www/MISP# mysql -u $MYSQL_USER -p$MYSQL_PASSWORD -P $MYSQL_PORT -h $MYSQL_HOST -r -N $MYSQL_DATABASE
Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 37873
Server version: 11.8.4-MariaDB-log managed by https://aws.amazon.com/rds/

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [misp]> ^DBye
```

did you mean when set to false?
in that case something like this?

```
if [[ "${MYSQL_TLS}" != true ]]; then
  MYSQL_CMD+=" --skip-ssl"
fi
```

UFOSmuggler:
> @iglocska @oivindoh @UFOSmuggler @claudex @mortezamirkar please test this new version (start from a blank `.env` file)

have given everything a good look over on a fresh misp using this branch and i can't see anything of concern.

core/files/entrypoint.sh
ostefano: The problem is that when TLS is disabled, and trixie is used, the command `mysql` fails because TLS is enforced by the client unless `--skip-ssl` is used.

UFOSmuggler: ok cool, will sort it out.  i only tested with tls.

UFOSmuggler: see #337

mortezamirkar:
> @iglocska @oivindoh @UFOSmuggler @claudex @mortezamirkar please test this new version (start from a blank `.env` file)

I checked the code and enhance the build in #339

monjiapawne:
> Upgraded `misp-guard` as well. @monjiapawne please test this.

Tested with a clean .env and my preexisting config, all ran as expected. No issues with `misp-guard` on trixie as far as I can tell.

claudex:
I checked with a clean env, no issue.

PR #339, opened 2025-11-26
Build enhancement

mortezamirkar:
For new version in Trixie in #333  , i tried to enhance the build and add some customization to build image with full package or in slim mode
so first i try to create two group in bake file and separate full build and slim build and using inherits to reuse repeated parts
finally i update the readme and explain that how to build slim and full images


docker-bake.hcl
```diff
@@ -146,13 +165,14 @@ target "misp-core" {
...
+
   }
-  platforms = "${PLATFORMS}"
+  
 }
 
 target "misp-core-slim" {
-  context = "core/."
-  dockerfile = "Dockerfile"
+  inherits = [ "misp-core" ]
```
ostefano, line 175: With this 'inherits' can we remove all the vars definition from here that are in misp-core as well?

docker-bake.hcl
```diff
@@ -100,6 +104,21 @@ group "default" {
...
 }
 
+group "debian-slim" {
+  targets = [
+    "misp-modules-slim",
+    "misp-core-slim",
+    "misp-guard",
+  ]
+}
+group "debain" {
```
ostefano: Typo

ostefano (changes requested):
If you have cycles, it might be time to make the Dockerfile consistent with "slim/full" values instead of "lite/full"

README.md
```diff
@@ -238,6 +238,50 @@ For Okta, create a new application integration:
...
+
+Docker Buildx bake provides advanced build capabilities including multi-platform builds and parallel building of multiple targets. This method uses the `docker-bake.hcl` configuration file.
+
+**Prerequisites:**
+- Docker Buildx plugin installed and enabled
+- Properly configured `.env` file (copy from `template.env`)
+
+**Build full-featured images:**
+```bash
+docker buildx bake debian
```
ostefano, line 263: I believe you are forgetting this step:
```sed -e '/^[[:space:]]*$/d' -e '/[#@]/d' -e 's/\"//g' -e 's/\(^[^=]*\)=\(.*\)/\1="\2"/' template.env > env.hcl```

See build workflow

**Review by ostefano — CHANGES_REQUESTED** [review]

File: docker-bake.hcl (hunk 2 above)
Comment (mortezamirkar) [reply docker-bake.hcl]: done

File: docker-bake.hcl (hunk 1 above)
Comment (mortezamirkar) [reply docker-bake.hcl:175]: Yes, but the vars  definition in misp core and misp-core-slim should be same but in our case its same and there is a difference in MODULES_FLAVOR

Comment (ostefano) [reply docker-bake.hcl:175]: Sorry, I meant the args. Can we omit those the PYPI one and leave only the *_FLAVOR?

File: README.md (hunk 3 above)
Comment (mortezamirkar) [reply README.md:263]: tnx, i fixed this

**Comment by mortezamirkar** [comment]
> If you have cycles, it might be time to make the Dockerfile consistent with "slim/full" values instead of "lite/full"

Good idea, please create a issue related to this and i will do this at first possible time

File: README.md [hunk 4]
```diff
@@ -238,6 +238,54 @@ For Okta, create a new application integration:
...
+```
+
+This builds lightweight versions of `misp-core-slim`, `misp-modules-slim`, and `misp-guard` with reduced dependencies.
+
+**Available bake targets:**
+- `debian` - Full-featured images (misp-core, misp-modules, misp-guard)
+- `debian-slim` - Lightweight images (misp-core-slim, misp-modules-slim, misp-guard)
+- `default` - Builds all variants (both full and slim)
+
+**Note:** The `sed` command converts `template.env` to `env.hcl` format by removing empty lines, comments, and properly formatting variables for the bake file.
```
ostefano, line 281: Note that this command does not work on OSX, and folks should use GNU sed (aka gsed)

ostefano:
Done: https://github.com/MISP/misp-docker/issues/341

docker-bake.hcl
mortezamirkar: I think with some tricks it could be possible, iam gonna check this.

PR #342, opened 2025-12-01
Enhance core Dockerfile and Update build type 

### LINKED ISSUE #341: Update Dockerfile build type values for consistency
In the env we use slim/full.
In the Dockerfiles we use lite/full.

We should normalize them.
I am also open to change full to something else.

CC @mortezamirkar

mortezamirkar:
#341 
I Enhance the docker file and then update the build type with standard and slim
@oivindoh please test this and give me your result


ostefano:
The size of the diff (caused by ident changes) makes reviewing difficult.

Either of these would make reviewing easier.
- Split the commit in 'feature' and 'change ident'
or
- Describe in detail where the change of behaviors are introduced

mortezamirkar:
> The size of the diff (caused by ident changes) makes reviewing difficult.
> 
> Either of these would make reviewing easier.
> 
>     * Split the commit in 'feature' and 'change ident'
>       or
> 
>     * Describe in detail where the change of behaviors are introduced

Terminology standardization: Renamed lite → slim and full → standard throughout (Dockerfile, docker-bake.hcl, docker-compose.yml, template.env) for consistency with common Docker naming conventions.

Build optimization: Consolidated package installation by moving all apt-get install commands from the final stage up to the php-base stage. This reduces layer duplication and image size but doesn't change what gets installed.

Code formatting: Improved readability by adjusting indentation and comment alignment (no logic changes).

ostefano:
LGTM

mortezamirkar:
@ostefano I test my changes and that was ok, could you please merge this repo?

ostefano:
Will do as soon as I find time to test it 👍

mortezamirkar:
> Will do as soon as I find time to test it 👍

Thanks

ostefano (commented):
@mortezamirkar my main concern is that we move all packages to one layer. Would it be better instead to leave only those that are actually needed by each layer?

mortezamirkar:
@ostefano 
you are absolutely right ,that would be more optimal from a Docker best practices perspective.
However, I consolidated the packages because:

Both the slim and standard flavors need the exact same base dependencies . The only difference between them is the PHP extensions and specialized libraries , which are still conditionally installed based on CORE_FLAVOR.

Putting the PHP dependencies in one layer makes development and debugging of the code easier – when troubleshooting build issues or modifying dependencies, having them all in a single, early stage means fewer rebuilds and clearer dependency visibility.

ostefano (approved):

kiwi-33:
@mortezamirkar @ostefano This appears to break multi-arch docker builds.  I am using the `latest` tag in a dev environment and had to pin to an earlier version. https://github.com/MISP/misp-docker/issues/349

PR #343, opened 2025-12-02
#318 separate email config elements

### LINKED ISSUE #318: Separate ADMIN_EMAIL and MISP_EMAIL/MISP_CONTACT
Currently, the `ADMIN_EMAIL` environment variable is used to configure both the administrator (user id=1) email and the email used for sending GPG-signed notifications.
This limits flexibility, as there is no way to specify different email addresses for these purposes.

We propose adding support for separate environment variables, `MISP_EMAIL` (or `MISP_CONTACT`)

Currently, these parameters must be manually changed after each restart.

UFOSmuggler:
Fixes #318.

Without the `MISP_CONTACT` and `MISP_EMAIL` envars being set, they default to `ADMIN_EMAIL` so there should be no breaking behavior.

When set, `MISP_EMAIL` is used for gpg, and `MISP_CONTACT` is used as contact email address.  `ADMIN_EMAIL` remains the envar to define the email address/username of user 1.

With these two set in `.env`:

```
MISP_CONTACT=info@admin.test
MISP_EMAIL=gpg@admin.test
```

We see:

```
# grep -E "('contact|'email)" configs/config.php
    'contact' => 'info@admin.test',
    'email' => 'gpg@admin.test',
    'email' => 'gpg@admin.test', # <---- this one is in the gnupg part of the config file

# gpg --homedir gnupg --list-keys --keyid-format=long
gpg: checking the trustdb
gpg: marginals needed: 3  completes needed: 1  trust model: pgp
gpg: depth: 0  valid:   1  signed:   0  trust: 0-, 0q, 0n, 0m, 0f, 1u
/home/ufosmuggler/stuff/misp-docker/gnupg/pubring.kbx
------------------------------------------------
pub   rsa3072/9CFC7DE1364E13F6 2025-12-02 [SCEAR]
      B0FBF2A231C972B2D2F1356A9CFC7DE1364E13F6
uid                 [ultimate] MISP Admin <gpg@admin.test>
```

While changing these envar values will change the config file, the gnupg keyring will not be automatically regenerated as this could constitute data loss.  This would be an exercise for the operator.


docker-compose.yml
```diff
@@ -144,6 +144,8 @@ services:
       - "DISABLE_PRINTING_PLAINTEXT_CREDENTIALS=${DISABLE_PRINTING_PLAINTEXT_CREDENTIALS}"
       # standard settings
       - "ADMIN_EMAIL=${ADMIN_EMAIL}"
+      - "MISP_CONTACT=${MISP_CONTACT}"
+      - "MISP_EMAIL=${MISP_EMAIL}"
```
ostefano, line 148: Can you document these two settings inside template.env?
Could we just the name into MISP_CONTACT_EMAIL, and MISP_GPG_EMAIL?
So it is clear what they are used for.

ostefano (changes requested):

docker-compose.yml
UFOSmuggler: > Can you document these two settings inside template.env? Could we just the name into MISP_CONTACT_EMAIL, and MISP_GPG_EMAIL? So it is clear what they are used for.

they are actually [already in template.env](https://github.com/MISP/misp-docker/blob/master/template.env#L138) which is why you saw nothing for that file in my commit:

```
# These variables allows overriding some MISP email values.
# They all default to ADMIN_EMAIL.

# MISP.email, used for notifications. Also used
# for GnuPG.email and GPG autogeneration.
# MISP_EMAIL=

# MISP.contact, the e-mail address that
# MISP should include as a contact address
# for the instance's support team.
# MISP_CONTACT=
```

how they were used was slightly off (missing from docker-compose.yml, so don't make it into the container env), otherwise they would have already worked.  my changes fix that, and also tidy them up to work in the same style we generally do the environment variables.

regarding changing the naming, they directly reflect the misp config item names, so i think we should leave them as-is:

| Name | Desc |
| - | - |
| MISP.email | The e-mail address that MISP should use for all notifications |
| MISP.contact | The e-mail address that MISP should include as a contact address for the instance's support team. |

ostefano (approved):

PR #359, opened 2026-01-02
Switch to hardened python images

### LINKED ISSUE #357: Switch to hardened python images
Hello, 

everything is in the title. Could it be a good idea to switch to these hardened python docker images or not? or just only core image (ui front) ?
=> https://hub.docker.com/hardened-images/catalog/dhi/python/images
=> https://hub.docker.com/hardened-images/catalog/dhi/valkey
=> https://hub.docker.com/hardened-images/catalog/dhi/mysql

Image

ostefano:


ostefano:
@arteta22000 mmm, registry requires login.

Not sure whether we want to force all users to do so.

arteta22000:
> @arteta22000 mmm, registry requires login.
> 
> Not sure whether we want to force all users to do so.


Ummm, if I understand correctly, the Docker images are hosted on GitHub?
So you only need authentication to the Docker registry (DHI) for the pipelines that build the images, right?
In that case, you can just use a token hub docker in the pipelines to build the images.
All users will continue to use the GitHub Docker registry in the docker compose

ostefano:
> > @arteta22000 mmm, registry requires login.
> > Not sure whether we want to force all users to do so.
> 
> Ummm, if I understand correctly, the Docker images are hosted on GitHub? So you only need authentication to the Docker registry (DHI) for the pipelines that build the images, right? In that case, you can just use a token hub docker in the pipelines to build the images. All users will continue to use the GitHub Docker registry in the docker compose

Correct, and documentation for users building the images (for compliance reasons, there are more than some).

Also, see my other concerns detailed here https://github.com/MISP/misp-docker/issues/357#issuecomment-3705268530

ostefano:
@arteta22000 correction: if we change also valkey and myself, everybody doing `docker compose up` will need a `docker` account.

PR #370, opened 2026-01-26
Allow for a registry mirror

firefart:
docker supports `registry-mirror` only for Dockerhub images and not for other domains / registries. If you use a package mirror for other registries, you have to change the image name from `ghcr.io/xxxx` to `registry-mirror/ghcr.io/xxxx`. To avoid having to change the original docker-compose.yml or override it with a `.override.yml` file and keep the versions in sync, this change adds a `REGISTRY_MIRROR_URL` env variable to specify an optional registry mirror. If not defined it defaults to an empty string resulting in the original image name.

Example:

Without `REGISTRY_MIRROR_URL`:

```
ghcr.io/egos-tech/smtp:1.1.3
```

With `REGISTRY_MIRROR_URL=internal.mirror.com/`:

```
internal.mirror.com/ghcr.io/egos-tech/smtp:1.1.3
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
       - "SMARTHOST_ALIASES=${SMARTHOST_ALIASES}"
 
   redis:
-    image: valkey/valkey:7.2
+    image: ${REGISTRY_MIRROR_URL:-}valkey/valkey:7.2
     restart: always
     command: |
       sh -c '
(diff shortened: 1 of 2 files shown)


ostefano:
LGTM but should we extend this to all images?

ostefano (changes requested):
Same trick should work for DockerHub images.

firefart:
@ostefano sure, included it for the dockerhub urls too

ostefano (approved):

PR #372, opened 2026-02-04
Set default value for OIDC_AUTH_METHOD

deade1e:
Since OIDC_AUTH_METHOD is not checked for not being set, and therefore gives no error if it's missing, simply give it a reasonable default value

### DIFF
--- a/core/files/configure_misp.sh
+++ b/core/files/configure_misp.sh
@@ -100,7 +100,7 @@ set_up_oidc() {
                 \"role_mapper\": ${OIDC_ROLES_MAPPING},
                 \"default_org\": \"${OIDC_DEFAULT_ORG}\",
                 \"mixedAuth\": ${OIDC_MIXEDAUTH},
-                \"authentication_method\": \"${OIDC_AUTH_METHOD}\",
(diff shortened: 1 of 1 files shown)


core/files/configure_misp.sh
```diff
@@ -100,7 +100,7 @@ set_up_oidc() {
                 \"role_mapper\": ${OIDC_ROLES_MAPPING},
                 \"default_org\": \"${OIDC_DEFAULT_ORG}\",
                 \"mixedAuth\": ${OIDC_MIXEDAUTH},
-                \"authentication_method\": \"${OIDC_AUTH_METHOD}\",
+                \"authentication_method\": \"${OIDC_AUTH_METHOD:-client_secret_post}\",
```
ostefano, line 103: Can we do this check before? I.e., where all other checks are done (main entrypoint file)?

ostefano (changes requested):

core/files/configure_misp.sh
deade1e: Sure: https://github.com/MISP/misp-docker/pull/373

Closing this.

PR #375, opened 2026-02-14
Fix Timezone Handling for Containers

LSI-ZuagrastaWastl:
This PR standardizes timezone handling across all MISP Docker containers.

The timezone can now be set optionally via the `TZ` environment variable.
If defined, it is applied globally and consistently across all containers and all processes, including system services, supervisord, background jobs, logs, and the MISP web interface.

If `TZ` is not set, the default remains `UTC`.

### Changes
- Added TZ as environment variable with default `UTC`
- Mounted /usr/share/zoneinfo/${TZ:-UTC} to /etc/localtime
- Configured PHP (tzone.ini) to use date.timezone from TZ

### Benefits

- Consistent timezone across system, logs, workers, and web interface
- Optional but globally enforced configuration
- More portable Docker setup
- Non-breaking change

Best Regards from Bayern-CERT


docker-compose.yml
```diff
@@ -362,6 +372,12 @@ services:
       retries: 3
       start_period: 10s
 
+configs:
```
ostefano: Could you move this to a separate file and edit Dockerfile to COPY it at build time and run sed in there (like we do with the PHP files?

Your solution is much cleaner, but not everybody is using `docker-compose.yml`.

ostefano (changes requested):

docker-compose.yml
LSI-ZuagrastaWastl: In fact, I had primarily focused on the fact that when using `docker-compose.yml`, you have the option of setting the time zone everywhere.  And as far as I can see, `configs:` must therefore remain in `docker-compose.yml` for the PHP-Settings.

As far as I can see, for the build-time is the timezone setting via `${PHP_TIMEZONE}`.
https://github.com/MISP/misp-docker/blob/b12a20c37885e98466969ec0cf63ec53f9df9d0a/core/files/entrypoint_fpm.sh#L50
Therefore, in my opinion, no additional file would be necessary.
However, this variable has no effect on a finished image via `docker-Compose.yml.`

I could imagine that you could ensure that either `${PHP_TIMEZONE}` or` ${TZ:-UTC}` is used. This means you only have to set it once in `.env` in example like this 
`sed -i "s/;?date.timezone = .*/date.timezone = ${PHP_TIMEZONE:-${TZ:-UTC}}/" "$FILE"`

And probably here 
https://github.com/MISP/misp-docker/blob/b12a20c37885e98466969ec0cf63ec53f9df9d0a/core/Dockerfile#L116
would be better `ENV TZ= ${TZ:-UTC}`.

What do you think about it @ostefano ?

LSI-ZuagrastaWastl:
@ostefano 
I took another closer look at everything and found a few errors. `PHP_TIMEZONE` could never be set as default in `.env` and was never set due to syntax errors in the entrypoint_fpm.sh. I fixed the error and tied everything to TZ.

core/files/entrypoint.sh
```diff
@@ -101,6 +101,9 @@ export SUPERVISOR_HOST=${SUPERVISOR_HOST:-127.0.0.1}
 export SUPERVISOR_USERNAME=${SUPERVISOR_USERNAME:-supervisor}
 export SUPERVISOR_PASSWORD=${SUPERVISOR_PASSWORD:-supervisor}
 
+# Setting Timezone for supervisord
+ln -sf /usr/share/zoneinfo/${TZ:-UTC} /etc/localtime
```
ostefano: Should we use debian specific commands for this instead?

ostefano (changes requested):

core/files/entrypoint.sh
LSI-ZuagrastaWastl: You're right. With `update-alternatives --install /etc/localtime localtime /usr/share/zoneinfo/${TZ} 0` , it's probably cleaner. I will change it.

ostefano (changes requested):
Sometimes we use `${TZ:UTC}` and sometimes `${TZ:-UTC}`, why?
Do we still need the PHP_TIMEZONE var, can't we just use TZ?

LSI-ZuagrastaWastl:
The missing `-` was my mistake. Without it, UTC would not be used as the default.
Since PHP_Timezone was never set via `.env`, I had already considered this. However, I wasn't sure at first whether this variable might still be used elsewhere. But that's not the case. I have now converted PHP_Timezone to TZ or removed it.

Now everything should be fine, in my opinion.

core/files/entrypoint.sh
```diff
@@ -102,6 +103,11 @@ export SUPERVISOR_HOST=${SUPERVISOR_HOST:-127.0.0.1}
 export SUPERVISOR_USERNAME=${SUPERVISOR_USERNAME:-supervisor}
 export SUPERVISOR_PASSWORD=${SUPERVISOR_PASSWORD:-supervisor}
 
+# Setting Timezone for supervisord
+update-alternatives --install /etc/localtime localtime /usr/share/zoneinfo/${TZ:-UTC} 0
```
ostefano: Why do we always load the var with a fallback value, if we do it just a few lines above?

Maybe 107/109 should use a non-interpolated TZ var?

ostefano (changes requested):
Just one last nit

core/files/entrypoint.sh
LSI-ZuagrastaWastl: Yes, you're right. That shouldn't and mustn't be the case here. I'll change it. Thanks for pointing that out.

LSI-ZuagrastaWastl: I also took another look at the other files. However, the fallback is necessary again in the compose file.

ostefano: That makes sense. My concern was limited to this instance only.

ostefano (approved):

ostefano (approved):

PR #389, opened 2026-03-11
Add UUID aka MISP.uuid envar

UFOSmuggler:
Like #385 it turns out that `MISP.uuid` is also used for [session handling](https://github.com/MISP/MISP/blob/a83d47ea4f0b8b1cc509bb471b87877ff4b23a86/app/Controller/AppController.php#L1321) and will cause CSRF issues when diverged across instances in a load balanced configuration.

This implements the same treatment as the `SALT` envar.  Default behaviour does not change:

1. If envar and config value is blank, a MISP.uuid is autogenerated
2. If envar is set, will set MISP.uuid to whatever is specified
3. If envar is blank but config value is not, config value is respected

From what I can tell, this is the last setting used in session handling that impacts HA deployments.

I have also added some advice to the `README.md` on using the image in HA deployments.


ostefano:
I wonder if in the long term we should have just one var and a JSON encoded dictionary inside that var.

UFOSmuggler:
> I wonder if in the long term we should have just one var and a JSON encoded dictionary inside that var.

i'll have a think on this

UFOSmuggler:
i have some ideas around constructing and passing in a single envar but i have some big ticket work items to punch out first (tip migration)

maybe in a few weeks to a month.

Issue #387, opened 2026-03-09
Need help with configuring LDAPAUTH correctly in kubernetes deployment


Macleykun:
Hello!

I'm using Kustomize and i've been trying to configure LDAPauth but no matter how i try to configure it, it seems to break the application and in the logging and checking the config file i noticed the following:

```Provided new config is not array, `NULL` given.```

In the /var/www/MISP/app/Config/config.php i can see other variables i've set, like the redis_password, but when i go to the ldapauth part, i do not see that part.

This is the redacted version of what i'm using now, adding it into the instance-secrets.env file:

```
LDAPAUTH_ENABLE=true
LDAPAUTH_LDAPSERVER=ldaps://domain.ext
LDAPAUTH_LDAPDN=DC=domain,DC=ext
LDAPAUTH_LDAPREADERUSER=CN=test-ldap,OU=Service Accounts,OU=management,DC=domain,DC=ext
LDAPAUTH_LDAPREADERPASSWORD=;Hdfsd3466M8k+K346Y,Bdsafasdf63346]$h?cADj.Gsdss346afaf'[ifasdfdasf79WnS[asda436sdf#T?Xsdaf=BJ.]asdfsafnj;{?
LDAPAUTH_LDAPSEARCHFILTER=(memberof:1.2.840.113556.1.4.1941:=CN=group-misp,OU=Application ext,OU=Groepen,DC=domain,DC=ext)
LDAPAUTH_LDAPSEARCHATTRIBUTE=mail
LDAPAUTH_LDAPEMAILFIELD=mail
LDAPAUTH_LDAPNETWORKTIMEOUT=-1
LDAPAUTH_UPDATEUSER=true
LDAPAUTH_LDAPDEFAULTORGID=1
LDAPAUTH_LDAPDEFAULTROLEID=3
LDAPAUTH_DEBUG=false
LDAPAUTH_LDAPTLSCUSTOMCACERT=/usr/local/share/ca-certificates/rootca.crt
LDAPAUTH_LDAPPROTOCOL=3
LDAPAUTH_LDAPALLOWREFERRALS=true
LDAPAUTH_STARTTLS=false
LDAPAUTH_MIXEDAUTH=true
LDAPAUTH_LDAPTLSREQUIRECERT=LDAP_OPT_X_TLS_NEVER
LDAPAUTH_LDAPTLSCRLCHECK=LDAP_OPT_X_TLS_CRL_PEER
[...]

PR #392, opened 2026-03-11
Add ldapauth example

Macleykun:
As requested here https://github.com/MISP/misp-docker/issues/387#issuecomment-4039429633

Here's the PR with my example config and explanation what it exactly does. This way users understand the importance of:

- not using quotes.
- not use a trailing slash.
- not worry about escaping.
- using a recursive ldap filter
- using proper tls (ldaps)

If there's any questions or adjustments you would like to see, lmk!

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,43 @@ You can configure LDAP authentication in MISP using 2 methods:
 
 LdapAuth is recommended over ApacheSecureAuth because it doesn't require rproxy apache with the ldap module.
 
+The following configuration to use LdapAuth plugin with your ldap/AD server has been tested and hardend.
+
+This example expects that you mounted the rootca under `/usr/local/share/ca-certificates/rootca.crt` into your pod, which is automatically added to the `/etc/ssl/certs/ca-certificates.crt` and `/etc/ssl/certs/ca-certificates.crt` bundle.
+
+The use of `memberOf:1.2.840.113556.1.4.1941:=` makes the group `MIPS-ALLOW-IN` which is a <u>member of</u> security-role group, allows the <u>members</u> in the security-role to be authorized (RBAC).
+
+This example uses `userPrincipalName` attribute as the username the user logs into, and `mail` to send emails to when the user has logged in.
+
+You don't need to use any quote's, double quote's, escape or double escape, the /configure_misp.sh script makes sure to single quote what needs quote's into `/var/www/MISP/app/Config/config.php`.
+
+Make sure to have this defined into the `instance-secrets.env`.
+
+```bash
+BASE_URL=https://misp.apps.openshift.domain.local
+LDAPAUTH_ENABLE=true
+LDAPAUTH_LDAPSERVER=ldaps://domain.local
+LDAPAUTH_LDAPDN=OU=Company,OU=Management,DC=domain,DC=local
+LDAPAUTH_LDAPREADERUSER=CN=ldap-account,OU=Accounts LDAP,OU=Management,DC=domain,DC=local
+LDAPAUTH_LDAPREADERPASSWORD=YoucanType4nythingH3r3Even1'Are3scaped!
+LDAPAUTH_LDAPSEARCHFILTER=(&(objectCategory=person)(objectClass=user)(memberOf:1.2.840.113556.1.4.1941:=CN=MIPS-ALLOW-IN,OU=Groups Applications,OU=Groups,DC=domain,DC=local))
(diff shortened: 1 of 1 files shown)


ostefano (changes requested):
I've tried to rewrite it a bit (but I still do not understand the the `memberOf` sentence. Can you rewrite it?

```
LdapAuth is recommended over ApacheSecureAuth because it doesn't require rproxy apache with the ldap module.

The following configuration to use LdapAuth plugin with your LDAP/AD server has been tested and hardend.

This example expects rootca mounted under `/usr/local/share/ca-certificates/rootca.crt` into your pod, automatically added to the `/etc/ssl/certs/ca-certificates.crt` and `/etc/ssl/certs/ca-certificates.crt` bundle.

The use of `memberOf:1.2.840.113556.1.4.1941:=` makes the group `MIPS-ALLOW-IN` which is a <u>member of</u> security-role group, allows the <u>members</u> in the security-role to be authorized (RBAC).

This example uses the `userPrincipalName` attribute as the MISP username, and the `mail` attribute as the email address used for notification when the user logs in.

You do not need to use any quotes, double quotes, escapes or double escapes, `/configure_misp.sh` will take care of it before copying the value inside `/var/www/MISP/app/Config/config.php`.

Make sure to have this defined inside your environment (`.env` file).

```bash
BASE_URL=https://misp.apps.openshift.domain.local
LDAPAUTH_ENABLE=true
LDAPAUTH_LDAPSERVER=ldaps://domain.local
LDAPAUTH_LDAPDN=OU=Company,OU=Management,DC=domain,DC=local
LDAPAUTH_LDAPREADERUSER=CN=ldap-account,OU=Accounts LDAP,OU=Management,DC=domain,DC=local
LDAPAUTH_LDAPREADERPASSWORD=YoucanType4nythingH3r3Even1'Are3scaped!
LDAPAUTH_LDAPSEARCHFILTER=(&(objectCategory=person)(objectClass=user)(memberOf:1.2.840.113556.1.4.1941:=CN=MIPS-ALLOW-IN,OU=Groups Applications,OU=Groups,DC=domain,DC=local))
LDAPAUTH_LDAPSEARCHATTRIBUTE=userPrincipalName
LDAPAUTH_LDAPEMAILFIELD=mail
LDAPAUTH_LDAPNETWORKTIMEOUT=-1
LDAPAUTH_UPDATEUSER=true
LDAPAUTH_LDAPDEFAULTORGID=1
LDAPAUTH_LDAPDEFAULTROLEID=3
LDAPAUTH_DEBUG=false
LDAPAUTH_LDAPTLSCUSTOMCACERT=false
LDAPAUTH_LDAPPROTOCOL=3
LDAPAUTH_LDAPALLOWREFERRALS=false
LDAPAUTH_STARTTLS=false
LDAPAUTH_MIXEDAUTH=true
LDAPAUTH_LDAPTLSREQUIRECERT=LDAP_OPT_X_TLS_DEMAND
LDAPAUTH_LDAPTLSCRLCHECK=LDAP_OPT_X_TLS_CRL_PEER
LDAPAUTH_LDAPTLSPROTOCOLMIN=LDAP_OPT_X_TLS_PROTOCOL_TLS1_2
```

Note: LDAPAUTH_STARTTLS is set to false as it caused errors in the log (this should be investigated further).
```

Macleykun:
Hi! Hows the latest change?
Also i believe the starttls errors as it's already hardcoded in the config to use LDAPS. Therefore upgrading a unencrypted connection to an encrypted (or vise versa), isn't desired.

We do have hardening in place where we can't use LDAP to do search queries aswell, so that could also be a reason why i get an error but still properly can connect.

Also, while i think this example is already long:
`LDAPAUTH_LDAPSEARCHFILTER=(&(objectCategory=person)(objectClass=user)(memberOf:1.2.840.113556.1.4.1941:=CN=MIPS-ALLOW-IN,OU=Groups Applications,OU=Groups,DC=domain,DC=local))`

Users can also make an OR statement to add more different groups to allow in. Ofcours permissions/roles have to be manually set by an admin.

`LDAPAUTH_LDAPSEARCHFILTER=(&(objectCategory=person)(objectClass=user)(|(memberOf:1.2.840.113556.1.4.1941:=CN=MIPS-ALLOW-IN,OU=Groups Applications,OU=Groups,DC=domain,DC=local)(memberOf:1.2.840.113556.1.4.1941:=CN=MIPS-ALLOW-IN-ASWELL,OU=Groups Applications,OU=Groups,DC=domain,DC=local)))`

But i don't want to make the example to broad, as that might confuse people.

ostefano (approved):

Issue #398, opened 2026-03-20
PHP-Timezone sed pattern fails on whitespace variance


LSI-ZuagrastaWastl:
Unfortunately, I only just found out that the sed pattern for `php.ini` timezone is too whitespace-specific. 

The problem existed from the start, and I had fixed it right away in my initial tests, but unfortunately didn’t include the fix in the final PR.

Fix: Remove the space from pattern (`= .*` → `=.*`) to make it flexible.

PR coming soon.

PR #399, opened 2026-03-20
Remove space from php-timezone sed pattern

### LINKED ISSUE #393: Bug when setting TZ to Europe/Amsterdam
See discussion here https://github.com/MISP/misp-docker/commit/aee58c4320cd833de636883da4de66daff703a3f#commitcomment-179696132

LSI-ZuagrastaWastl:
Remove the space from the pattern (`= .*` → `=.*`) to make it more flexible.

Solution for #398 

sorry for my mistake :/


core/files/entrypoint_fpm.sh
```diff
@@ -47,7 +47,7 @@ change_php_vars() {
         sed -i "s/session.sid_length = .*/session.sid_length = 64/" "$FILE"
         sed -i "s/session.use_strict_mode = .*/session.use_strict_mode = 1/" "$FILE"
         echo "Configure PHP | Setting 'date.timezone = ${TZ}'"
-        sed -i "s|^;date.timezone = .*|date.timezone = ${TZ}|" "$FILE"
```
ostefano, line 50: Why do we need to fix it?

My understanding is that, if this is never modified manually, it should work consistently?

LSI-ZuagrastaWastl: The problem is that, by default, the `php.ini` file does not appear to have a space after `;date.timezone =`. As a result, the pattern does not recognize this line and does not set the value.

Without the space in the pattern, it is therefore more robust.

ostefano: Oh, that is indeed annoying...

ostefano (approved):

Issue #400, opened 2026-03-23
Not all supervisord use TZ


LSI-ZuagrastaWastl:
During further testing, I discovered that not all Supervisors actually adopt the TZ timezone.
As far as I could tell, the cause is `echo $TZ /etc/locatime`. I can’t quite figure out why this is the case. This command was actually intended to serve as an additional safeguard.
However, without this command, all Supervisors use the TZ value

https://github.com/MISP/misp-docker/blob/a4e7a00221ea13ef03f8f7d985063a54b6e756e5/core/files/entrypoint.sh#L109

PR #401, opened 2026-03-23
Bug-fix for : Not all supervisord use TZ

LSI-ZuagrastaWastl:
Fixes issues: #400 

As far as I can tell, the cause is `echo $TZ /etc/locatime`. I don't quite understand why this is the case. This command was actually intended to serve as an additional safeguard.
However, without this command, all Supervisors use the TZ value.


core/files/entrypoint.sh
```diff
@@ -105,8 +105,6 @@ export SUPERVISOR_PASSWORD=${SUPERVISOR_PASSWORD:-supervisor}
 
 # Setting Timezone for supervisord
 update-alternatives --install /etc/localtime localtime /usr/share/zoneinfo/${TZ} 0
-# Also set /etc/timezone for broader compatibility
-echo "${TZ}" > /etc/timezone
```
ostefano, line 109: what about `timedatectl set-timezone` ?

LSI-ZuagrastaWastl: No. It makes the same mistakes :/
It seems that it's better not to explicitly set the time zone in `/etc/timezone`

ostefano: That is quite sub-optimal because there is no validation :/

ostefano (approved):

PR #403, opened 2026-03-26
Allow to disable the automatic rdb snapshot in valkey

claudex:
By default, Valkey dumps regularly a snapshot, on testing/demo environnment, we may want to avoid the extra fs pressure

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,7 +23,7 @@ services:
     restart: always
     command: |
       sh -c '
-        if [ "$${ENABLE_REDIS_SNAPSHOT}" = "false" ] ; then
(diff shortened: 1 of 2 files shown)


ostefano (changes requested):
Should it be "DISABLE" since this is soemthing that should always be enabled bar additional cases?

claudex:
Yes, but I find less confusing to read something `ENABLE=true` than `DISABLE=false` because it avoids double negation. But I can change it if you prefer.

ostefano:
My point is that if you provide an option called ENABLE_X, I would expect X to be disabled by default.

claudex:
I've renamed the option.

ostefano (approved):

PR #404, opened 2026-03-31
Allow setting MISP.curl_request_timeout via env var

firefart:
This allows setting the `MISP.curl_request_timeout` setting via the `CURL_REQUEST_TIMEOUT` environment variable


ostefano:
@UFOSmuggler we really need a better approach here :) (clarification: here = MISP settings via env var)

firefart:
@ostefano should I look into a mechanism to set arbitrary settings to address this in a more general way?

ostefano:
Yup, at least for all the options ending up in MISP I believe we should have just one env variable, with all values encoded inside JSON encoded dictionary. Not a trivial change.

firefart:
@ostefano https://github.com/MISP/misp-docker/pull/405

PR #405, opened 2026-03-31
allow setting custom settings

firefart:
This allows for custom MISP settings at startup. Mounting a json file is way easier than putting a escaped json in the .env file

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,23 @@ If you push a change to add or remove an environment variable, please look in "c
 
 The misp-core container has definitions for minimum safe default settings which are set if needed each time the container starts. They will only be set if there is no existing entry in the config.php file or database for these settings. If you specify a custom value for any of these settings it will be respected. See the definitions of these in "core/files/etc/misp-docker" where the filenames contain the word "defaults".
 
+#### Setting custom settings
+
+If you want to set custom MISP settings on startup please take the following steps:
+
+- create a `settings.json` file in the project root
+- uncomment the `settings.json` mountpoint in the `misp-core` image in `docker-compose.yml`
+- if you change the default mount path, be sure to update `MISP_SETTINGS_FILE` in your `.env` file
+- add the settings to your json
+
+##### Example
+
+```json
+{
+  "MISP.curl_request_timeout": "600"
+}
+```
+
 #### Storing system settings in the DB
 
 This container includes the "ENABLE_DB_SETTINGS" environment variable, which can be used to set "MISP.system_setting_db" to true or false. This changes the behaviour of where MISP chooses to store operator made settings changes; in config.php or in the system_settings database table. By default this is set to false.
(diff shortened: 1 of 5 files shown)


ostefano:
@firefart thank you for this, but I would want to spend a bit more time on it, and have @UFOSmuggler collaborate as well. We would also need a tool to convert existing configurations

firefart:
The file is empty to begin with, so existing installations are not affected. If people want to migrate existing configurations to this setup they simply need to put in their settings line by line, other settings are untouched. So I guess we need no tool to convert existing configurations, or is there already something else in place to set custom configuration settings that needs to be migrated?

ostefano:
There are some catches. Some settings requires the force flag when set via the `setSetting` command, and there are sets of settings that need to be added first (because they rely on others) so it's not super easy sailing. That's why I would like @UFOSmuggler to give it a look first.

ostefano:
Alternatively if you feel taking up the challenge see how the existing settings are grouped together (optional, minimum, etc), and try to replicate it. Maybe we can leave the JSON as-is and pass it to each section for "additiona" non-env var specified settings?

firefart:
Any update on this @UFOSmuggler ?

UFOSmuggler:
working on it.  have something that preserves the current way things work, and also allows all config to occur in a single json file, or a mix of envars, safe defaults, and json.  substantial restructure of how it all works again, will let us get rid of a lot of envars, and should make it trivial/unnecessary to add new stuff, but allows us to put guard rails on things that need it.

Kagee:
An alternativ way to change the settings (assuming file-based settings are used) is to give the (validated) JSON file as input to `MISP/tests/modify_config.php` a file that replaces the old config file with the new settings from JSON, and outputs the old file. We do something simmilar to this:
```
sudo -u www-data php /var/www/MISP/tests/modify_config.php modify "$(cat /opt/misp/data/config-patch.json)" > "/var/www/MISP/app/Config/config.php-$(date +%s).bak"
```

firefart:
Any updates on this?

ostefano:
@UFOSmuggler ?

@firefart can you rebase in the meanwhile?

README.md
```diff
@@ -100,6 +100,23 @@ If you push a change to add or remove an environment variable, please look in "c
...
+- create a `settings.json` file in the project root
+- uncomment the `settings.json` mountpoint in the `misp-core` image in `docker-compose.yml`
+- if you change the default mount path, be sure to update `MISP_SETTINGS_FILE` in your `.env` file
+- add the settings to your json
+
+##### Example
+
+```json
+{
+  "MISP.curl_request_timeout": "600"
```
ostefano, line 116: Is `600` being a string intended?

Is it because `MISP.curl_request_timeout` does not accept integers?

docker-compose.yml
```diff
@@ -327,6 +329,8 @@ services:
       - "PHP_SESSION_COOKIE_SAMESITE=${PHP_SESSION_COOKIE_SAMESITE:-Lax}"
       # compose profiles
       - "COMPOSE_PROFILES=${COMPOSE_PROFILES}"
+      # MISP settings
+      - "MISP_SETTINGS_FILE=${MISP_SETTINGS_FILE:-/settings.json}"
```
ostefano, line 333: Can you set the default inside the main entrypoint as well?

ostefano (changes requested):

PR #410, opened 2026-04-30
Add /var/www/MISP/app/Console to PATH

cyberdashy:
- /var/www/MISP/app/Console contains the cake CLI used for admin tasks
- Without it in PATH,  every docker exec admin command requires the full path
- Simple quality of life improvement


ostefano:
LGTM

ostefano (approved):

PR #414, opened 2026-05-20
fix: set date.timezone in PHP CLI php.ini

### LINKED ISSUE #413: PHP CLI timezone not set – MISP workers run in wrong timezone
## Summary

`entrypoint_fpm.sh` sets `date.timezone` only in `/etc/php/*/fpm/php.ini`, but not in `/etc/php/*/cli/php.ini`.

MISP background workers (e.g. CakePHP console commands) use the CLI SAPI and therefore ignore the FPM timezone config. As a result, all worker-generated timestamps use the container's default UTC regardless of the `TZ` environment variable.

I didn't notice this during my previous checks and tests, since I was manually triggering pulls and pushes, for example. However, it appears that the set time zone is indeed being used in those cases.

## Steps to reproduce

1. Set `TZ` to a time zone other than UTC, e.g.: `TZ=Europe/Berlin` in the container environment.
2. Take a look in `logs/misp-workers.log`
3. Observe that the logged time is UTC, not Europe/Berlin.

## Expected behaviour

`date.timezone` in both `fpm/php.ini` and `cli/php.ini` reflects the `TZ` env var.

## Fix

Add the following loop to `change_php_vars()` in `entrypoint_fpm.sh`:

```bash
for FILE in /etc/php/*/cli/php.ini; do
    [[ -e $FILE ]] || break
    sed -i "s|^;date.timezone =.*|date.timezone = ${TZ}|" "$FILE"
done
```

LSI-ZuagrastaWastl:
## Problem

`change_php_vars()` only patched `date.timezone` in FPM's `php.ini`.
PHP-CLI (used by MISP workers/console commands) kept the default UTC timezone, ignoring the `TZ` env var.
MISP background workers (e.g. CakePHP console commands) use the CLI SAPI and therefore ignore the FPM timezone config. As a result, all worker-generated timestamps use the container's default UTC regardless of the TZ environment variable.

## Change

Added a loop to `entrypoint_fpm.sh` that applies `date.timezone = ${TZ}` to all `/etc/php/*/cli/php.ini` files, consistent with the existing FPM handling.

## Testing

1. Set `TZ` to a time zone other than `UTC`, e.g.: `TZ=Europe/Berlin` in the container environment.
2. Take a look in `logs/misp-workers.log`
3. Observe that the logged time is `Europe/Berlin`.

Fixes #413 

_P.S. I thought creating a separate entrypoint script, such as `entrypoint_cli.sh`, would be a bit too much overhead for a single change. If you ever want to modify or set more PHP CLI settings, this one will be moved along with them._


ostefano:
LGTM

ostefano (approved):

PR #417, opened 2026-06-03
chore: cleanup VirtioFS workarounds, parameterize PHP version, add Makefile, and fix minor issues

mortezamirkar:
<dl>
<dt>1. safe_sed_i() helper + VirtioFS cleanup</dt>
<dd>Added a <code>safe_sed_i()</code> function using <code>mktemp+cat</code> to replace the repetitive <code>sed > tmp; cat tmp > file; rm tmp</code> workaround for VirtioFS (Docker Desktop for Mac). Replaced <code>dd if=... of=...</code> copies with <code>cat ... > ...</code> for clarity.</dd>
<dt>2. Parameterize hardcoded PHP 8.4</dt>
<dd>Added <code>PHP_PACKAGE_VERSION</code> build arg (default <code>8.4</code>) to <code>core/Dockerfile</code>. Replaced 31 hardcoded references across Dockerfile, nginx config, and entrypoint scripts. Scripts now use version-agnostic <code>php-fpm</code> command via build-time symlink. Nginx config uses <code>@PHP_VERSION@</code> placeholder templated at build time.</dd>
<dt>3. Add Makefile</dt>
<dd>New <code>Makefile</code> with <code>build</code>, <code>build-core</code>, <code>build-modules</code>, <code>build-guard</code>, <code>build-slim</code>, <code>bake</code>, <code>up</code>, <code>down</code>, <code>logs</code>, <code>shellcheck</code>, <code>clean</code>, and <code>prune</code> targets. <code>help</code> is the default.</dd>
<dt>4. Fix trailing comma in docker-bake.hcl</dt>
<dd>Added missing trailing comma after <code>GUARD_COMMIT</code> in the <code>misp-guard</code> target args block.</dd>
<dt>5. Remove duplicate AUTOCONF_ADMIN_KEY var</dt>
<dd>Removed <code>AUTOCONF_ADMIN_KEY</code> which was only a fallback for <code>AUTOGEN_ADMIN_KEY</code>. Simplified to a single <code>AUTOGEN_ADMIN_KEY</code> (default <code>true</code>).</dd>
</dl>


core/files/entrypoint.sh
```diff
@@ -51,8 +51,7 @@ export MISP_MODULES_FQDN=${MISP_MODULES_FQDN:-http://misp-modules}
 export ATTACHMENTS_DIR=${ATTACHMENTS_DIR:-/var/www/MISP/app/files}
 
 export AUTOCONF_GPG=${AUTOCONF_GPG:-true}
-export AUTOCONF_ADMIN_KEY=${AUTOCONF_ADMIN_KEY:-true}
-export AUTOGEN_ADMIN_KEY=${AUTOGEN_ADMIN_KEY:-$AUTOCONF_ADMIN_KEY}
```
ostefano, line 55: There was a reason for this. Can you check old commits and messages?

docker-bake.hcl
```diff
@@ -203,7 +203,7 @@ target "misp-guard" {
   tags = flatten(["${NAMESPACE}/misp-guard:latest", "${NAMESPACE}/misp-guard:${COMMIT_HASH}", GUARD_TAG != "" ? ["${NAMESPACE}/misp-guard:${GUARD_TAG}"] : []])
   args = {
     "GUARD_TAG": "${GUARD_TAG}",
-    "GUARD_COMMIT": "${GUARD_COMMIT}"
+    "GUARD_COMMIT": "${GUARD_COMMIT}",
```
ostefano, line 206: Nice catch!

ostefano (changes requested):
I like this thanks!

Just one comment that might require some spelunking (either git blame or github issues)

ostefano:
You could also split this PR, so we could merge the rest right now (and ADMIN_KEY issues postponed).

core/files/entrypoint.sh
mortezamirkar: Thanks for the review! You're right about AUTOCONF_ADMIN_KEY — it was kept as a backward compatibility alias so old .env files using the original variable name still work. I'll revert that commit

mortezamirkar:
@ostefano  Fixed! I've replaced the two-line pattern with a single export that keeps AUTOCONF_ADMIN_KEY as an inline fallback:
export AUTOGEN_ADMIN_KEY=${AUTOGEN_ADMIN_KEY:-${AUTOCONF_ADMIN_KEY:-true}}
Old .env files using AUTOCONF_ADMIN_KEY still work, but we no longer export the redundant variable into the environment. Everything else is unchanged — should be ready for merge.

ostefano (approved):

ostefano:
Superseded by https://github.com/MISP/misp-docker/pull/421

Changes:
- rebased
- made build var consistent

PR #423, opened 2026-06-10
Issue building with podman - Dockerfile ARG must be repeated in each FROM section

trolldbois:
Dockerfile ARG must be repeated in each FROM section
Without repeating ARG, the download of the composer.json file fails, due to CORE_TAG not being defined in that build section.


ostefano:
LGTM, can you fix also the other two images?

ostefano (changes requested):
See comment, would be nice if you could extend the fix to the other two images

ostefano:
Fixed in another branch

PR #432, opened 2026-07-16
Add OpenShift compatibility: rootless, arbitrary-UID support

jfstenuit:
MISP's images assumed a fixed UID and root-then-drop-privilege startup (`sudo`, `chown`, binding privileged ports, php-fpm's `user/group/listen.owner/listen.group` directives), which conflicts with OpenShift's `restricted` SCC - pods run under a randomly assigned, non-predictable UID with primary GID `0`, never root, and never with `CAP_CHOWN`/`CAP_SETUID`.

This PR reworks `misp-core`, `misp-modules`, and `misp-guard` to start as their final identity instead of dropping into it, following the standard "support arbitrary user IDs" container pattern (owner + group `0`, `g=u` permissions, setgid on directories so new files stay group-writable regardless of the runtime UID). Docker Compose, the raw Kubernetes manifests, and the Helm chart are all updated to match - including moving nginx off privileged ports 80/443, removing hardcoded `runAsUser`/`fsGroup` from pod security contexts, and generating all credentials (admin, API key, supervisord, MariaDB) as strong random secrets instead of static defaults. Non-OpenShift usage (plain Docker/Compose/vanilla k8s with a fixed UID) is unaffected - the default UID (`33`/`www-data`) is kept and simply added to supplementary group `0`.


ostefano:
Thanks for this.

We might need to give priority to this though at this stage https://github.com/MISP/misp-docker/pull/430 

Would a rebase possible (some of the concerns should already be addressed by that PR) once we merge?

jfstenuit:
There is ~40% overlap with the work of PR #430 . Probably better of create a new PR once #430 has been merged.

ostefano:
> There is ~40% overlap with the work of PR #430 . Probably better of create a new PR once #430 has been merged.

Could you do a review of that PR and check whether the direction is compatible with yours?

jfstenuit:
> > There is ~40% overlap with the work of PR #430 . Probably better of create a new PR once #430 has been merged.
> 
> Could you do a review of that PR and check whether the direction is compatible with yours?

Clearly, this goes in the right direction. The new nginx container is natively Openshift-compatible and the move to unprivileged ports aligns with the security requirements. After that change, I'll still have to adapt the init scripts and remove any leftover "sudo" or "chown" along the way. There is also the "-R" flag of php-fpm that needs to be dropped.

Also, the whole helm chart needs to be reviewed and aligned with the new Dockerfile.

I suggest I start a new branch once #430 is merged, with two change sets : 
- one to fix the remaining privilege issues in the common code
- one for the helm chart

pascaliske:
@jfstenuit Nice! I had the `core` container on my todo list as well after #430 is merged. Thanks for taking over. 😁 

In addition to the permission and UID changes I planned to switch the FPM / PHP config to a template based syntax as the nginx config in #430 instead of all the sed commands. Maybe you're interested in integrating that in your PR too?

<details>

<summary>PHP templates</summary>

```diff
# docker-compose.yml
services:
  misp-core:
    tmpfs:
      - /opt/php:mode=0755,uid=33,gid=33
```

```Dockerfile
# core/Dockerfile
ENV PHP_INI_SCAN_DIR="/etc/php/${PHP_PACKAGE_VERSION}/conf.d:/opt/php/conf.d"

...

# overwrite global php-fpm config to load dynamic pool configs
COPY <<-EOF /etc/php/${PHP_PACKAGE_VERSION}/fpm/php-fpm.conf
[global]
pid = /run/php/php${PHP_PACKAGE_VERSION}-fpm.pid
error_log = /var/log/php${PHP_PACKAGE_VERSION}-fpm.log
include=/etc/php/${PHP_PACKAGE_VERSION}/fpm/pool.d/*.conf
include=/opt/php/pool.d/*.conf
EOF
```

```shell
# core/files/entrypoint_fpm.sh
generate_php_config() {
    local REDIS_HOST_WITH_SCHEME
    local REDIS_PASSWORD_ESCAPED
    local STATUS_LISTEN

    mkdir -p /opt/php/{pool.d,conf.d}

    REDIS_HOST_WITH_SCHEME=$(echo "$REDIS_HOST" | grep -E '^\w+://' || echo "tcp://$REDIS_HOST")
    REDIS_PASSWORD_ESCAPED=$(printf '%s\n' "$REDIS_PASSWORD" | sed -e 's/[\/&]/\\&/g')

    if [[ "$ENABLE_REDIS_EMPTY_PASSWORD" == "true" ]]; then
        REDIS_SAVE_PATH="${REDIS_HOST_WITH_SCHEME}:${REDIS_PORT}"
    elif [[ -n "$REDIS_PASSWORD" ]]; then
        REDIS_SAVE_PATH="${REDIS_HOST_WITH_SCHEME}:${REDIS_PORT}?auth=${REDIS_PASSWORD_ESCAPED}"
    else
        echo "ERROR: REDIS_PASSWORD is not set but ENABLE_REDIS_EMPTY_PASSWORD is false. Please set REDIS_PASSWORD or enable ENABLE_REDIS_EMPTY_PASSWORD=true for passwordless Redis."
        exit 1
    fi

    # FPM listen address
    if [[ -n "$PHP_LISTEN_FPM" ]]; then
        FPM_LISTEN=$([[ "$DISABLE_IPV6" == "true" ]] && echo "0.0.0.0:9002" || echo "[::]:9002")
    fi

    # FPM status block
    if [[ -n "$FASTCGI_LISTEN_STATUS" ]]; then
        STATUS_LISTEN=$([[ "$DISABLE_IPV6" == "true" ]] && echo "0.0.0.0:9003" || echo "[::]:9003")
        FPM_STATUS_BLOCK="pm.status_listen=${STATUS_LISTEN}"$'\n'"pm.status_path=/fpm-status"
    else
        FPM_STATUS_BLOCK=""
    fi

    # pm.max_spare_servers: must be >= pm.start_servers
    if [[ "$PHP_FCGI_START_SERVERS" -gt "$PHP_FCGI_SPARE_SERVERS" ]]; then
        PHP_FCGI_MAX_SPARE_SERVERS="$PHP_FCGI_START_SERVERS"
    else
        PHP_FCGI_MAX_SPARE_SERVERS="$PHP_FCGI_SPARE_SERVERS"
    fi

    # --- render templates ---
    export REDIS_SAVE_PATH FPM_LISTEN FPM_STATUS_BLOCK PHP_FCGI_MAX_SPARE_SERVERS
    envsubst < /opt/php/templates/99-runtime.ini > /opt/php/conf.d/99-runtime.ini
    envsubst < /opt/php/templates/99-www.conf > /opt/php/pool.d/99-www.conf
}
```

```ini
# core/files/opt/php/templates/99-runtime.ini
memory_limit=${PHP_MEMORY_LIMIT}
max_execution_time=${PHP_MAX_EXECUTION_TIME}
upload_max_filesize=${PHP_UPLOAD_MAX_FILESIZE}
max_file_uploads=${PHP_MAX_FILE_UPLOADS}
post_max_size=${PHP_POST_MAX_SIZE}
max_input_time=${PHP_MAX_INPUT_TIME}
date.timezone=${TZ}

session.save_handler=redis
session.save_path='${REDIS_SAVE_PATH}'
session.sid_length=64
session.use_strict_mode=1
```

```ini
# core/files/opt/php/templates/99-www.conf
listen=${FPM_LISTEN}
pm.max_children=${PHP_FCGI_CHILDREN}
pm.start_servers=${PHP_FCGI_START_SERVERS}
pm.min_spare_servers=${PHP_FCGI_SPARE_SERVERS}
pm.max_spare_servers=${PHP_FCGI_MAX_SPARE_SERVERS}
pm.max_requests=${PHP_FCGI_MAX_REQUESTS}
${FPM_STATUS_BLOCK}
```

</details>

t-desmond:
Really appreciate all the work that's gone into this! I tried deploying the chart myself and ran into a couple of issues. I managed to fix most of them, but after that the pod still ends up in a crash loop because:

* `customize_misp.sh` restarts the workers (`EACCES`, exit 244)
* `entrypoint_nginx.sh` restarts `php-fpm`, which then fails due to the read-only filesystem and a socket conflict
* `nginx` exits, and the liveness probe eventually kills the pod

At this point, I think I'll just sit quietly and wait for the fix to be merged. 😅 Thanks again for all the work you're putting into this!

ostefano:
@jfstenuit this should be now rebased, otherwise I will close it

ostefano (changes requested):
Please rebase

ostefano:
@jfstenuit we merged the other big change (nginx container) so this might be a good time to revisit this.

Let me know if you are interested, otherwise I will close this.

jfstenuit:
Hi @ostefano . Pending time re-allocation, but rebase and adaptation is definitely foreseen within the month.

jfstenuit:
PR is now re-based, but still require some tests : 

- effective deployment on OpenShift
- effective deployment on docker (to ensure no regression introduced)

README.md
```diff
@@ -55,6 +55,7 @@ docker compose version
 
 1. Copy the `template.env` file to `.env` in the project root.
 2. Customize `.env` according to your requirements (optional but recommended).
+3. `make init-dirs` (or manually `mkdir -p configs logs files ssl gnupg && chown -R 33:0 configs logs files ssl gnupg`) - `misp-core` runs as a non-root user and can no longer `chown` these bind-mounted host directories itself on first start.
```
ostefano: Does the execution of misp-core fails if those dirs are not created / chowned corrrectly? We should fail early if possible

ostefano (changes requested):

README.md
pascaliske: I would assume that misp-core fails because the permissions of those dirs will be wrong on first start. But this could probably be automated using init containers (for both docker and k8s):

- Docker compose: https://docs.docker.com/compose/how-tos/init-containers/#fix-volume-ownership-before-a-non-root-service-starts

PR #444, opened 2026-09-09
Prevent ADMIN_KEY envar from recreating existing auth key

### LINKED ISSUE #443: ADMIN_KEY envar expires and recreates matching keys on startup without preserving allow lists
If the ADMIN_KEY envar is set, the following code block in configure_misp will run on every container start and potentially expire then recreate matching keys without preserving any allowlist entries set for the admin key.

```
    if [ -n "$ADMIN_KEY" ]; then
        if [ "$DISABLE_PRINTING_PLAINTEXT_CREDENTIALS" == "true" ]; then
            echo "... setting admin key from environment variable"
        else
            echo "... setting admin key to '${ADMIN_KEY}'"
        fi
        CHANGE_CMD=(sudo -u www-data /var/www/MISP/app/Console/cake User change_authkey 1 "${ADMIN_KEY}")
```

This should be updated to check if the user 1 key is already set to ADMIN_KEY, and if so, take no action to prevent creating duplicate entries or removing IP allow lists.

jeremiah-RENISAC:
Fixes #443

Adds a check when the environment variable "ADMIN_KEY" is set to see if user 1's key is already set to the same value by hash match, and if so, skips the change_authkey command. When change_authkey is run, it expires the current key and recreates a new duplicate entry without preserving any allowed ips.

### DIFF
--- a/core/files/configure_misp.sh
+++ b/core/files/configure_misp.sh
@@ -491,11 +491,11 @@ init_user() {
     fi
     elif [ -z "$ADMIN_KEY" ] && [ "$AUTOGEN_ADMIN_KEY" == "true" ]; then
         HAS_VALID_KEY=$($MYSQL_CMD -N -s -e 'SELECT EXISTS(SELECT 1 FROM auth_keys WHERE user_id = 1 AND (expiration = 0 OR expiration > UNIX_TIMESTAMP()));')
-	if (( HAS_VALID_KEY == 0 )); then
+	    if (( HAS_VALID_KEY == 0 )); then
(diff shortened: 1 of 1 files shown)


jeremiah-RENISAC:
Tested the new image:
- Redeploy with same ADMIN_KEY value triggers the "configured admin key already active, not changing" path
- Changing the ADMIN_KEY value and redeploying triggers the "setting admin key from environment variable" path

ostefano:
@righel can you help reviewing this?

